PR #3, opened 2017-09-12
Performance improvements for loading time & bug fixes for animations

vogelsgesang:
This pull request contains 3 types of changes: refactoring, performance improvements & fixes to animations.
The different changes are separate commits, but all part of the same pull request since they are based on each other.

### Refactoring

The very long callback function for the load event was split into few, smaller functions. Made it easier for me to grasp how the scripts work

### Performance

This pull request fixes two issues:
* The whole tree was drawn once with all nodes expanded. Immediately afterwards, unnecessary nodes where hidden. Especially for very complex graphs with a lot of nodes which are hidden by default, this lead to unnecessarily long loading times. For the dint7.xml example, this change brings the initial rendering time down from 600-650ms to 160-180ms and makes the entry animation much smoother.
* The `inline` loading logic relied on a previous network request to fail, first. Removing that network request saved between 10 and 20ms load time

### Animations

Various changes:
* The tree root enters directly at its final position and does not move during the entry animation
* For orientations bottom-to-top and right-to-left, the links between nodes were animated in from the wrong opposite side of the canvas. Now they are animated in from the correct side
* The exit animation of all nodes for orientations != top-to-bottom were going in the wrong direction.
* The tree canvas is positioned at bit higher.  So far, it was positioned slightly too low, so that the tree root was not visible for orientation bottom-to-top.


vogelsgesang:
Good point. Will take a look at the issues found by the linter

vogelsgesang:
Fixed the linter issues

d3/query-graphs.css
```diff
@@ -6,7 +6,6 @@ html, body {
 }
 #tree-container {
   padding: 0px;
-  padding-top: 53px;
```
rickcole: The original motivation for the relatively large top padding was to make space for the hover text of the root node. Without it the root hover text for all of the test queries are a little obscured and most of the root hover text is obscured for the dint queries, e.g., dint4.xml.

I suppose we could make the padding a function of orientation? Even better would be to make hover text always render visibly regardless of padding, but that may be more than we want to take on. 

Just noticed that the spinner is now also chopped off. Challenging to observe because of the fast layout and rendering! Try http://localhost:3000/d3/query-graphs.html?file=hyper-steps2.json&orientation=left-to-right&collapse=n. So this is not only a hover text issue.

vogelsgesang: I didn't notice that regression, thanks for catching it :)
I adapted the last commit such that:
* The root node is positioned a bit lower now, so that the tooltip is visible by default (adjusted orientations["top-to-bottom"].rooty)
* changes CSS such that the spinner is vertically centered

rickcole: Changes look good. Because we no longer have a result node root, which naturally had a small hover text, any non-trivial root node hover text will likely be partially obscured, but that was also already true for large hover text even with a result node. Someday we can improve the location awareness of hove text.

PR #6, opened 2017-10-25
Better labels for ref nodes

fgamador:


d3/query-graphs.js
```diff
@@ -451,6 +459,16 @@ function prepareTreeData(_treeData, convertHyPer) {
             }
         }
 
+        function handleBinding(node) {
+            if (node.properties && node.properties.name) {
+                node.name = node.properties.name;
+            } else if (node.properties && node.properties.ref) {
+                node.name = node.properties.ref;
+            } else {
+                node.name = "";
```
rickcole: Typically we fall back to node.tag rather than empty string, just to get some/any context for the node for debugging.

PR #15, opened 2017-12-07
Improve plan rendering for Hyper

vogelsgesang:
Since the visualization for Hyper was initially written, a few
properties were added to Hyper's query plans. These additional
informations cluttered the tree since they were rendered as nodes even
if they only added small additional knownledge and had no influence on
the overall shape of the query plan.

This commit changes the loading of Hyper query plans in the following
way:
* The `analyze` property is always displayed in the tooltip.
* Unknown properties are added to the tooltip if they are plain strings,
  numbers, booleans or null. Only if they are more complex, they are
  added to tree as child nodes. Using this heuristic, we will hopefully
  not have to adjust the loader again if Hyper adds new properties to
  the query graph.

----

Same query plan, nodes collapsed by default 

Before:
screen shot 2017-12-08 at 00 02 02

After:
screen shot 2017-12-08 at 00 02 48

### DIFF
--- a/d3/hyper.js
+++ b/d3/hyper.js
@@ -30,21 +30,35 @@ Miscellaneous
 
 var common = require('./common');
 
+// Convert to string. Return undefined if not supported.
+function toString(d) {
+    if (typeof (d) === "string") {
+        return d;
+    } else if (typeof (d) === "number") {
+        return d.toString();
+    } else if (typeof (d) === "boolean") {
+        return d.toString();
+    } else if (d === null) {
+        return "null";
+    } else if (d === undefined) {
+        return "undefined";
+    }
+    return undefined;
+}
+
+// Convert to string. Returns the JSON serialization if not supported.
+function forceToString(d) {
+    var str = toString(d);
+    if (str === undefined) {
+        str = JSON.stringify(d);
+    }
+    return str;
+}
+
 // Convert Hyper JSON to a D3 tree
 function convertHyper(node, tag) {
     var innerNode;
     if (typeof (node) === "object" && !Array.isArray(node)) {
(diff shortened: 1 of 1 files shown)


d3/hyper.js
```diff
@@ -30,159 +30,118 @@ Miscellaneous
 
 var common = require('./common');
 
+// Convert to string. Return undefined if not supported.
```
luiseeo: kind of a Javascript question, does this function return undefined if no explicit return statement was called?

vogelsgesang: yes, `undefined` is the default return value if no value is returned.
Probably I should explicitly write `return undefined, though

vogelsgesang: modified the pull request to include `return undefined`

rickcole:
Just wondering. Why is the node count larger in the after picture, 4079 versus 3615? I would have expected the count of nodes after to be less than before because we would be created fewer nodes.

vogelsgesang:
The increased node number is due to multiple reasons:
* Some of the branches of the input tree were previously missing. E.g., for `nullif` we did not show the `valueForComparison`
* For expressions, `left` and `right` are now part separate nodes in the tree.

Screenshot of the expression `CAST(100*v435 AS double precision) / NULLIF(v529+v530, 0)`:

before this change:
screen shot 2017-12-09 at 16 28 12

after this change:
screen shot 2017-12-09 at 16 28 51

Note that before this change:
* the type for the cast was not displayed (also not part of the tooltip)
* the second parameter to nullif was missing
* in the new picture, the iurefs are lacking the IU names (something which I will address in a follow-up commit which cleans up the representation of expressions)

PR #16, opened 2017-12-07
Allow to show additional fields in the tree label

vogelsgesang:
The tree label (displayed in the top-right corner of the canvas)
contained a hardcoded set of infos, so far. With this commit, two
sources are used to propagate the information displayed in this label:
* Information can be added using the `properties` query parameter.
  This is useful, e.g., for Tableau Log Viewer such that it can augment
  the tree with meta information such as the current optimizer
  step/rewrite or the query execution time.
* The loaders can contribute information. E.g., the Hyper query
  plan loader uses this to display the number of returned columns.

### DIFF
--- a/d3/query-graphs.js
+++ b/d3/query-graphs.js
@@ -78,6 +78,23 @@ switch (graphCollapse) {
         break;
 }
 
+// Get properties to be rendered in the toplevel info card
+var toplevelProperties;
+if (queryObject.properties) {
+    try {
+        toplevelProperties = JSON.parse(queryObject.properties);
+    } catch (err) {
+        document.write("JSON parse failed with '" + err + "'.");
+    }
+} else {
+    toplevelProperties = {};
+}
+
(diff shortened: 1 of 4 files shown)


d3/query-graphs.js
```diff
@@ -292,10 +309,23 @@ function abbreviateNames(treeData) {
     }, common.allChildren);
 }
 
+//
+// Escapes a string for HTML
+//
+function escapeHtml(unsafe) {
```
luiseeo: I remember that this function showed up in another pull request. Can the function be moved to a common place?

vogelsgesang: Yes, it was part of the upload-server.js
I preferred not to introduce a dependency between client code and server code.
But if you want, I can move it to a common location

luiseeo: No, I think it is ok as it is. I think it is more important than the server code stays decoupled

PR #18, opened 2017-12-13
Reduce loading time by cutting down dependencies

vogelsgesang:
This pull request contains the following individual commits which greatly reduce the loading time of the visualizations, especially over slow VPNs. The number of bytes sent over the network is reduced from more than 2.5MB to approx. 500KB.

__Remove dependency on jquery from query-graph.js__

We used jquery only on 2 lines and it is straightforward to replace those two usages by native functions. By removing jquery, we significantly decrease the size of our bundle:
    * unminified, uncompressed size: 2.5MB -> 1.7MB
    * minified, uncompressed size: 407 KB -> 322 KB

__Remove jquery from static html pages__

__Minified sample_graph.png__

Using optipng, we were able to reduce file size from 40KB to 22KB

__Only use the parser from xml2js__

We did only use the parser from xml2js but nevertheless included the complete library. By only including the parser, we are able to reduce the size of our bundle:
    * unminified, uncompressed size: 1.9 MB -> 1.7 MB
    * minified, uncompressed size: 321.5 KB -> 275.3 KB

__Add compression midleware to upload-server__

For our unminified JS bundle the number of transfered bytes over the network was decreased from 1.6MB to 377KB


result-table/compare_layout.html
```diff
@@ -95,8 +93,8 @@ <h3 id="right-label">After rewrites</h3>
     function setLabels() {
         var leftLabel = getParameterByName("left-label") || "A";
         var rightLabel = getParameterByName("right-label") || "B";
-        $("#left-label").text(leftLabel);
-        $("#right-label").text(rightLabel);
+        document.getElementById("left-label").innerText=leftLabel;
```
luiseeo: nitpick: space between the assignment operator `innerText = leftLabel` vs `innerText=leftLabel`; same with the next line

vogelsgesang:
[caniuse.com](https://caniuse.com) looks good for both innerHTML and getElementById.
Also, the IE screenshots from https://netrenderer.com/ indicate that we are able to render the tree correctly in IE 9 & IE 10.
(To use netrenderer you need to make your local server accessible from the internet. I usually use [ngrok](https://ngrok.com) for that purpose)

PR #21, opened 2017-12-17
Fix edge labels for horizontal tree layouts

vogelsgesang:
Edge labels were broken for all layouts except top-to-bottom because
orientation.x/y were not called. Also, the node's x0/y0 were already
updated before using them for the enter-animation of the edge labels.
Thereby, the edge labels were animated in from the parent node's end
position instead of its position at the start of the animation.

In order to make sure that this type of mistakes are harder to make in
the future, I changed when we are switching between the logical and the
display coordinate space. Before this change node.x/y were refering to
logical coordinates while x0/y0 were in the display coordinate system.
Now, they are both in the display coordinate system. Thereby, it is no
longer to transform the coordinates using orienation.x()/y() for each
individual SVG element drawn.

### DIFF
--- a/d3/query-graphs.js
+++ b/d3/query-graphs.js
@@ -502,10 +502,13 @@ function drawQueryTree(treeData) {
         .size(ooo.size);
 
     // Define a d3 diagonal projection for use by the node paths later on.
-    var diagonalRaw = d3.svg.diagonal();
-    var diagonal = d3.svg.diagonal()
-        .projection(function(d) {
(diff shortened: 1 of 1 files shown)


vogelsgesang:
I just realized that without the `diagonal.projection` call, the tangents of the links are incorrect for horizontal layouts. I will keep this change around since it is useful for the upgrade to D3 version 4 (D3v4 does not support `diagonal.projection` anymore). But for now, it makes no sense to merge this

PR #28, opened 2018-06-06
Better collapsing for Hyper plans

vogelsgesang:
So far, the keys which should be collapsed in a Hyper plan were
hardcoded. Since a few (uncommon) operators like tablefunction and
explicitscan were not taken into account, those operators were
displayed poorly.

With this change, we no longer ship a hardcoded list of keys which
should be collapsed. Instead, we now:
* for expressions: collapse no children at all. This is the same
  behavior as before.
* for operators: collapse all children except for a few selected
  children such as "left" and "right".

With this change, we now display also uncommon operators nicely.
At the same time, we will not have to update the lists of collapsed
keys when Hyper devs add new functionality.

### DIFF
--- a/d3/hyper.js
+++ b/d3/hyper.js
@@ -23,7 +23,8 @@ function convertHyper(node, parentKey) {
         };
     } else if (typeof (node) === "object" && !Array.isArray(node)) {
         // "Object" nodes
-        var children = [];
+        var explicitChildren = [];
+        var additionalChildren = [];
         var properties = {};
 
         // Take the first present tagKey as the new tag. Add all others as properties
@@ -51,9 +52,9 @@ function convertHyper(node, parentKey) {
             }
             var child = convertHyper(node[key], key);
             if (Array.isArray(child)) {
-                children = children.concat(child);
+                explicitChildren = explicitChildren.concat(child);
             } else {
-                children.push(child);
+                explicitChildren.push(child);
             }
         });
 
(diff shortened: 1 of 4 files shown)


d3/hyper.js
```diff
@@ -85,16 +86,35 @@ function convertHyper(node, parentKey) {
...
+        } else {
+            // Everything else (usually expressions): display uncollapsed
+            children = explicitChildren.concat(additionalChildren);
+        }
         // Build the converted node
         var convertedNode = {
             tag: tag,
             properties: properties,
             children: children,
+            _children: _children,
```
luiseeo: what is the variable `_children` for?

vogelsgesang: `children` and `_children` are used for collapsing. Each click which expands/collapses a node internally only swaps `children` and `_children`. This means, `children` should contain the initially visible children while `_children` should contain all children, including the ones which are hidden by default

d3/hyper.js
```diff
@@ -185,51 +205,11 @@ function generateDisplayNames(treeData) {
...
         }
-    }, function(n) {
-        return n.children;
-    });
-}
-
-function collapseNodes(treeData, graphCollapse) {
-    var streamline = graphCollapse === "s" ? common.streamline : common.collapseChildren;
-    var collapseChildren = common.collapseChildren;
-    if (graphCollapse !== 'n') {
```
rickcole: Are we de-supporting the graphCollapse options for Hyper plans?

vogelsgesang: While I personally don't use the `graphCollapse`, I adjusted the commit so that we keep supporting `graphCollapse`

rickcole: Great!

Issue #25, opened 2017-12-21
edge labels misplaced for all orientations except top-to-bottom


rickcole:
See pull request #21 for details. Three examples of misplaced edge labels below. Orientation left-to-right places edge labels as if the layout was top-to-bottom, while bottom-to-top and right-to-left labels are somewhere not apparently visible on the display. 

http://localhost:3000/d3/query-graphs.html?file=hyper-query1.json&orientation=left-to-right
http://localhost:3000/d3/query-graphs.html?file=hyper-query1.json&orientation=right-to-left
http://localhost:3000/d3/query-graphs.html?file=hyper-query1.json&orientation=bottom-to-top

rickcole:
Fixed along with crosslinks support.

PR #30, opened 2018-08-22
Add crosslinks between nodes related by properties

rickcole:
Crosslinks are edges between nodes related by operator property
references, such as by the common table expression source property,
early probe builder property, and magic set magic property.

Two versions of crosslinks are drawn: not highlighted and highlighted.
Initial crosslinks are not highlighted, to avoid cluttering the existing
tree layout, while the highlighted version appears upon node mouseover.

Also fixed issue #25, misplaced edge labels.

### DIFF
--- a/d3/hyper.js
+++ b/d3/hyper.js
@@ -58,6 +58,24 @@ function convertHyper(node, parentKey) {
             }
         });
 
+        // Add the following keys as children only when they refer to objects and display as properties if not
+        var objectKeys = ["source"];
+        objectKeys.forEach(function(key) {
+            if (!node.hasOwnProperty(key)) {
+                return;
+            }
+            if (typeof node[key] !== "object") {
+                properties[key] = common.forceToString(node[key]);
+                return;
+            }
+            var child = convertHyper(node[key], key);
+            if (Array.isArray(child)) {
+                explicitChildren = explicitChildren.concat(child);
+            } else {
+                explicitChildren.push(child);
+            }
+        });
+
         // Display these properties always as properties, even if they are more complex
         var propertyKeys = ["analyze"];
         propertyKeys.forEach(function(key) {
@@ -68,7 +86,7 @@ function convertHyper(node, parentKey) {
         });
 
         // Display all other properties adaptively: simple expressions are displayed as properties, all others as part of the tree
-        var handledKeys = tagKeys.concat(childKeys, propertyKeys);
+        var handledKeys = tagKeys.concat(childKeys, objectKeys, propertyKeys);
         Object.getOwnPropertyNames(node).forEach(function(key, _index) {
             if (handledKeys.indexOf(key) !== -1) {
                 return;
@@ -89,6 +107,7 @@ function convertHyper(node, parentKey) {
             additionalChildren.push({tag: key, children: innerNodes});
             return;
         });
+
         // Display the cardinality on the links between the nodes
         var edgeLabel = node.hasOwnProperty("cardinality") ? common.formatMetric(node.cardinality) : undefined;
         // Collapse nodes as appropriate
@@ -165,6 +184,10 @@ function generateDisplayNames(treeData) {
                 node.name = node.tag;
                 node.symbol = 'table-symbol';
                 break;
+            case "explicitscan":
+                node.name = node.tag;
+                node.symbol = "temp-table-symbol";
+                break;
             case "temp":
                 node.name = node.tag;
                 node.symbol = "temp-table-symbol";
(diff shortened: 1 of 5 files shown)


d3/hyper.js
```diff
@@ -58,6 +58,24 @@ function convertHyper(node, parentKey) {
             }
         });
 
+        // Add the following keys as children only when they refer to objects
+        var objectKeys = ["source"];
```
vogelsgesang: why can't we simply add `source` to the `childKeys` array?

d3/hyper.js
```diff
@@ -208,6 +231,61 @@ function generateDisplayNames(treeData) {
     }, common.allChildren);
 }
 
+// Function to find a node with a given operatorId assuming left-deep assignment
```
vogelsgesang: I don't think we should be making the "left-deep assignment" assumption

d3/hyper.js
```diff
@@ -208,6 +231,61 @@ function generateDisplayNames(treeData) {
...
+            default:
+                if (node.hasOwnProperty("properties") && node.properties.hasOwnProperty("magic")) {
+                    sourcenodes.push({node: node, operatorId: node.properties.magic});
+                }
+                break;
+        }
+    }, common.allChildren);
+    // Add crosslinks from source to matching target node
+    sourcenodes.forEach(function(source) {
+        var targetnode = findNodeByOperatorId(root, source.operatorId);
```
vogelsgesang: Instead of calling findNodeByOperatorId multiple times and thereby traversing the tree multiple times, I would be in favor of creating a hash table for this. Something along the lines of:

```javascript
var operatorsById = {};
common.visit(root, function(node) {
   if (node.hasOwnProperty("properties") && node.properties.hasOwnProperty("operatorId")) {
      operatorsById[node.properties.operatorId] = node;
   }
}
```

and then this line here simply becomes
```javascript
var targetNode = operatorsById[source.operatorId];
```

we could even fill the `operatorsById` map as part of the tree traversal started in line 261 and wouldn't even need an additional traversal.

d3/query-graphs.js
```diff
@@ -770,19 +815,89 @@ function drawQueryTree(target, treeData) {
...
+                        y: source.y
+                    };
+                    return diagonal({
+                        source: o,
+                        target: o
+                    });
+                })
+                .remove();
+
+            // Update highlighted crosslinks
```
vogelsgesang: The code below seems to be duplicated with the code starting at line 839. The only difference I saw is the class name `crosslink` vs. `crosslink-highlighted`.
I think we should factor this into a helper function `updateCrosslinkPath(cssClass)` and then call it twice, once as `updateCrosslinkPath("crosslink")` and once as `updateCrosslinkPath("crosslink-highlighted")`

d3/query-graphs.js
```diff
@@ -592,6 +594,47 @@ function drawQueryTree(target, treeData) {
...
+    // Transition to unhighlight edges on mouseout
+    var edgeTransitionOut = function(path) {
+        path.transition()
+            .duration(duration)
+            .attr("opacity", 0)
+            .attr("d", diagonal);
+    };
+
+    // Handler builder for mouseover and mouseout
+    var mouseHandler = function(transition) {
```
vogelsgesang: this function is crosslink specific, but `mouseHandler` suggests a more general scope of this function.
Maybe rename to something like `crosslinkHighlightHandler`?
And also update the comment accordingly

d3/query-graphs.js
```diff
@@ -592,6 +594,47 @@ function drawQueryTree(target, treeData) {
...
+        return function(d) {
+            var crosslinks = svgGroup.selectAll("path.crosslink-highlighted");
+
+            // Filter the edges to those connected to the current node
+            crosslinks.filter(function(dd) {
+                var sourceMatches = d.x === dd.source.x && d.y === dd.source.y;
```
vogelsgesang: why are we matching based on the coordinates?
Can't we just match based on object identity, i.e. `d == dd.source`?

d3/query-graphs.js
```diff
@@ -592,6 +594,47 @@ function drawQueryTree(target, treeData) {
...
+    // Dash tween to make the highlighted edges animate from start node to end node
+    var tweenDash = function() {
+        var l = this.getTotalLength();
+        var i = d3.interpolateString("0," + l, l + "," + l);
+        return function(t) {
+            return i(t);
+        };
+    };
+
+    // Transition used to highlight edges on mouseover
```
vogelsgesang: Personally, I am not a huge fan of mouse-hover transitions.
They don't help me to understand the tree in an easier way but they force me to wait 0.75 seconds between hovering a link and seeing the result of that hover

But probably that is only my personal opinion, so feel free to ignore this comment

d3/query-graphs.js
```diff
@@ -592,6 +594,47 @@ function drawQueryTree(target, treeData) {
         update(d);
     }
 
+    // Dash tween to make the highlighted edges animate from start node to end node
+    var tweenDash = function() {
+        var l = this.getTotalLength();
+        var i = d3.interpolateString("0," + l, l + "," + l);
```
vogelsgesang: This gives are very nice visual effect in case I hover the target (i.e the `ExplicitScan`) since it will start drawing a solid line from the `ExplicitScan` to its corresponding `GroupBy`.

However, if I hover the `GroupBy` instead of the `ExplicitScan` the line is drawn towards the hovered `GroupBy` node and this feels a bit unnatural to me. Especially given that when hovering the `GroupBy`, I am mentally focusing on that `GroupBy` node and the solid line only arrives there ~0.75 seconds after the hover event

d3/query-graphs.js
vogelsgesang: Just to clarify: highlighting through mouse-hover is fine
I just don't like the animation delay

rickcole: I also find the delay annoying. Will likely keep some form of hover highlighting, but not delayed.

d3/query-graphs.js
rickcole: Will obviate difference by making hover highlight immediate, but then we lose directionality. Is direction, e.g., an ExplicitScan is a consumer, useful?

d3/query-graphs.js
rickcole: Indeed.

d3/query-graphs.js
rickcole: Agreed.

d3/query-graphs.js
rickcole: Yes. Also needs opacity.

d3/hyper.js
rickcole: Good suggestion that also facilitates crosslinks for earlier plans that did not have an operatorId defined for every node.

d3/hyper.js
rickcole: Agreed.

d3/hyper.js
rickcole: Both the `source` key and the `value` key (already in `childKeys`) are used for input objects and simple values, e.g., aggregations. There was no way to preserve existing behavior without special case handling of keys by name inside the `childKeys` iterator. Otherwise, all instances of `source` would be treated as children, adding additional verbosity.

d3/query-graphs.js
rickcole: Added flow directionality for crosslinks.

vogelsgesang: looks really nice! :)

PR #33, opened 2019-03-04
Hawkfish tql

hawkfish:
I think this is at a good stopping point.


d3/tql.js
```diff
@@ -0,0 +1,1002 @@
+// -----------------------------------------------------------------------------
+//
+// This file is the copyrighted property of Tableau Software and is protected
```
rickcole: The overall project license is MIT. This conflicting copyright should be removed.

package.json
```diff
@@ -5,7 +5,7 @@
   "description": "Helping people see and understand queries.",
   "main": "d3/query-graphs.js",
   "scripts": {
```
rickcole: Need to add d3/tql.js to list of "lint" source files. (Comment here because GitHub doesn't allow commenting the correct line as it's not in the PR.) Run lint `npm run lint`. May want to disable some of the errors/warnings for this file, but not all.

Also, please run `npm run bundle-tlv`, which will fail because the version of uglify we require is not ES6 and tql.js expects ES6 features. I tried using latest version of uglify-js@3, but then there are other complaints. It might be simplest to remove use of default function parameters, *groan*. Or, work through the other ES6 issues using uglify-js@3. Or, use a transpiler, e.g., babel.

d3/tql.js
hawkfish: Removed.

package.json
hawkfish: OK, I fixed the things that were real errors (`eqeqeq`) and suppressed all the complaints about my (more readable) use of spaces. If the file wasn't about 1000 lines with an error on every line I might have been convinced, but it was just too much work.

 I also tried to lint my unit tests, but `eslint` does not understand how `mocha` extends the namespace so I gave up. IMHO, unit tests are more important than format fascism!

README.md
```diff
@@ -21,8 +21,9 @@ View example query visualizations at
 * <https://tableau.github.io/query-graphs/d3/query-graphs.tlv.html?file=tableau/joins.xml>,
 * <https://tableau.github.io/query-graphs/d3/query-graphs.tlv.html?file=tableau/query-function.xml&collapse=n>,
 * <https://tableau.github.io/query-graphs/d3/query-graphs.tlv.html?file=tableau/dint4.xml>,
-* <https://tableau.github.io/query-graphs/d3/query-graphs.tlv.html?file=hyper/query2.json>, or
-* <https://tableau.github.io/query-graphs/d3/query-graphs.tlv.html?file=hyper/steps2.json&orientation=left-to-right>.
+* <https://tableau.github.io/query-graphs/d3/query-graphs.tlv.html?file=hyper/query2.json>,
+* <https://tableau.github.io/query-graphs/d3/query-graphs.tlv.html?file=hyper/steps2.json&orientation=left-to-right>, or
+* <https://tableau.github.io/query-graphs/d3/query-graphs.tlv.html?file=tql/iejoin.tql&collapse=n>
```
rickcole: If adding example, it would be nice to also add TQL to the list of query ecosystem artifacts listed above?

package.json
```diff
@@ -5,15 +5,15 @@
...
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "mocha --watch --recursive",
```
rickcole: Great!

PR #37, opened 2019-06-08
Split query-graphs core library and standalone-server into separate projects

vogelsgesang:
This PR splits the reusable library part from the website-specific
parts and puts them into two separate projects. Thereby this commit
turns the repo into a mono-repo hosting multiple projects with
independent `package.json` files.

The goal of this split is that the core `query-graphs` library can be
easily embedded into other environments (such as IDEs)

While at it, this commit also adjusts a few minor things:
* compare-layout.html contained a broken link
* added Richard Wesley & myself to the contributor list in `package.json`

Note: this PR alse includes #35 and #36 - rebasing/merging would be too painful


package.json
```diff
@@ -6,14 +6,14 @@
...
-    "bundle-list": "browserify d3/query-graphs.js --list",
+    "bundle": "browserify d3/app.js -d -o d3/query-graphs.bundle.js",
+    "bundle-list": "browserify d3/app.js --list",
     "prebundle-uglify": "npm run bundle",
     "bundle-uglify": "uglifyjs --compress --mangle -- d3/query-graphs.bundle.js > d3/query-graphs.min.js",
     "prebundle-tlv": "npm run bundle-uglify",
     "bundle-tlv": "tar -czf $npm_package_name-tlv-qt-$npm_package_version.tgz -C d3 query-graphs.tlv.html query-graphs.css query-graphs.min.js",
     "postinstall": "npm run bundle-tlv",
-    "lint": "eslint upload-server.js d3/query-graphs.js d3/common.js d3/colors.js d3/hyper.js d3/tableau.js d3/json.js d3/xml.js d3/tql.js"
+    "lint": "eslint upload-server.js d3/app.js d3/query-graphs.js d3/common.js d3/colors.js d3/hyper.js d3/tableau.js d3/json.js d3/xml.js d3/tql.js"
```
rickcole: missing rename to tree-rendering.js?

.gitignore
```diff
@@ -7,7 +7,8 @@
 .vscode
 npm-debug.log
 node_modules
-media/uploads/*
-!media/uploads/3d5c935c.xml
-d3/query-graphs.bundle.js
```
rickcole: removed bundle.js from ignore, but then changed query-graphs.html to reference min.js?

i think we stopped pushing all (?) build artifacts at some point, but then we needed min.js in the repo for github-pages, which uses query-graphs.tlv.html

so i think we need min.js, but not bundle.js?

standalone-server/webroot/query-graphs.html
```diff
@@ -25,10 +25,10 @@
...
     <meta name="theme-color" content="#ffffff">
     <!-- main style sheet  -->
-    <link href='query-graphs.css' rel='stylesheet' type='text/css'>
+    <link href='query-graphs.bundle.css' rel='stylesheet' type='text/css'>
 </head>
 <body>
 </body>
 <!-- bundle created by browserify -->
-<script src="query-graphs.bundle.js"></script>
+<script src="query-graphs.min.js"></script>
```
rickcole: any impact to debugability?

package.json
vogelsgesang: fixed.
Note how in the followup commit, this line is changed to `eslint .`, thereby including all files by default. The `.eslintignore` makes sure we don't lint the bundled and minified source

standalone-server/webroot/query-graphs.html
vogelsgesang: yes, this turned out to be a bit annoying. Reverted it

.gitignore
vogelsgesang: yes, you are right:
* we should keep checking in min.js for the github-pages
* bundles.js should remain in the .gitinore

Adjusted my commit

vogelsgesang:
Updated the PR such that:
* .gitignore keeps the bundle.js
* the latest min.js is included
* missing parameters are passed through correctly (`debug`, `orientation`)
* standalone-server no longer depends on d3. Instead it uses query-graphs' d3 dependency. `npm` couldn't deal with both packages depending on the same library.

PR #38, opened 2019-06-13
Upgrade to D3 version 5

vogelsgesang:
This PR upgrades our D3 dependency from version 3.5.16 to 5.9.2.

The original motivation of this was to reduce file size. Unfortunately, the impact turned out to be smaller than I hoped for. We only reduce file size of our minified JS bundle from 297 kB to 259 kB.

Still, I think this is worth it, and also it is nice to be up-to-date with our dependencies anyway.

The largest changes were:
* the tree layout now works on a hierarchy created through
  `d3.hierarchy`. This requires some plumbing to get this hierarchy work
  with our loaders.
  I did not want to introduce a dependency to `d3.hierarchy` directly
  into the loaders, hence all changes are done in `tree-rendering.js`
* `.enter().append()` no longer adds newly added elements to the default
  selection. This required some calls to `.merge` to retain the old
  behavior
* switching to d3-fetch: this means, older browsers won't be supported
  anymore. I think that should be fine, given https://caniuse.com/#search=fetch

I tested this in:
* Safari 11.0.3
* Firefox 67.0.2
* TLV
* Chrome 54

Doesn't work in IE. But IE didn't work before either.
Couldn't test in Edge since Edge refuses to open local IP addresses (not sure how to circumvent - don't care enough)


vogelsgesang:
I finally figured out how to take advantage of the ES6 modules of D3.
By changing from browserify to webpack and by using ES6 modules for our own code, too, I was able to reduce the size of our bundle to 230kB.
All in all we are at a ~22% size reduction now :)

rickcole:
Great. Will take a look on Monday.

On Jun 21, 2019, at 2:23 PM, Adrian Vogelsgesang <notifications@github.com<mailto:notifications@github.com>> wrote:


I finally figured out how to take advantage of the ES6 modules of D3.
By changing from browserify to webpack and by using ES6 modules for our own code, too, I was able to reduce the size of our bundle to 230kB.
All in all we are at a ~22% size reduction now :)

standalone-server/package.json
```diff
@@ -27,31 +27,33 @@
...
     "express": "^4.13.4",
     "fs": "0.0.2",
     "multer": "^1.1.0",
-    "parcelify": "^2.2.1",
     "query-graphs": "file:../query-graphs",
     "recursive-readdir": "^2.2.1",
     "spin": "0.0.1",
-    "uglify-js": "^2.7.0"
+    "style-loader": "^0.23.1",
+    "uglify-js": "^2.7.0",
```
rickcole: Not familiar with webpack: do we still need the explicit dependency on uglify-js or is webpack production mode enough?

standalone-server/webroot/query-graphs.html
```diff
@@ -24,11 +24,9 @@
     <meta name="msapplication-TileImage" content="icons/mstile-144x144.png">
     <meta name="msapplication-config" content="icons/browserconfig.xml">
     <meta name="theme-color" content="#ffffff">
-    <!-- main style sheet  -->
-    <link href='query-graphs.bundle.css' rel='stylesheet' type='text/css'>
 </head>
 <body>
 </body>
 <!-- bundle created by browserify -->
```
rickcole: Update comment (for webpack) or make generic or remove?

PR #39, opened 2019-08-07
More reusable core library

vogelsgesang:
This pull request bundles a few commits with the goal to make the core query-graphs library more reusable.

The largest changes are:
* Removing dependencies on the global `window` size
* Exposing control over the size of the drawing canvas
* Prefixing all our CSS classes with `qg-`. This is to avoid name clashes when loading our CSS into another app (e.g., into JupyterLab)

For more details see the individual commit messages


query-graphs/lib/hyper.js
```diff
@@ -307,3 +300,15 @@ export function loadHyperPlan(graphString, graphCollapse) {
     var crosslinks = addCrosslinks(root);
     return {root: root, crosslinks: crosslinks, properties: properties};
 }
+
+// Load a JSON tree from text
+export function loadHyperPlanFromText(graphString, graphCollapse) {
```
rickcole: Refactoring seems fine. Does it have any functional impact?

query-graphs/lib/json.js
```diff
@@ -39,13 +39,18 @@ function convertChildren(node) {
...
 
 // Load a JSON tree
-export function loadJson(graphString, _graphCollapse) {
+export function loadJson(json) {
+    var root = {name: "root", children: convertChildren(json)};
+    return {root: root};
+}
+
+// Load a JSON tree from text
+export function loadJsonFromText(graphString, _graphCollapse) {
```
rickcole: Refactoring seems fine. Does it have any functional impact?

query-graphs/lib/hyper.js
vogelsgesang: This refactoring here should not have any change in behavior for the "standalone-server".
When integrating query-graphs with JupyterLab, JupyterLab already provides the query plans as JSON. Hence, I needed a way to read plans from already parsed JSON and that's why I split the function into two.

query-graphs/lib/json.js
vogelsgesang: Same as above

PR #40, opened 2019-08-14
Jupyterlab integration

vogelsgesang:
This pull-request integrates QueryGraphs with JupyterLab.

query-graphs-jupyterlab-screenshot

It registers query-graphs as a renderer for the MIME type
`application/vnd.tableau.hyper-queryplan` with the associated file
extension `.plan.json`. Thereby, it enables two use cases:
1. Viewing `.plan.json` files stored in JupyterLab (e.g., uploaded by the user)
2. Visualizing plans sent by Jupyter kernels using the MIME type
    `application/vnd.tableau.hyper-queryplan`. Given a specialized
    kernel, this allows the direct visualization of query plans produced
    by `EXPLAIN VERBOSE`.

So far, no other query artifacts (Tableau's logical plans, TDE plans, ...)
are supported by this JupyterLab extension. We will add support for those
other plan types as the need arises.


vogelsgesang:
The compare-layout in the standalone-server still works as expected.
At first, I thought that thanks to JupyterLab's split-pane layouts, we would now have a strictly superior solution and hence removed the compare-layout.
However, I overlooked that the JupyterLab plugin does not yet expose the full set of functionality and is, e.g., missing support for XML query plans

vogelsgesang:
For now, I kept the compare-layout around and didn't remove it

PR #47, opened 2020-06-20
Modernize toolchain

vogelsgesang:
This series of commits modernizes our toolchain:
* All code is now written in TypeScript instead of JavaScript, thereby getting the benefits of static type checking. This commit does not yet actually make our code type-safe, it merely sets-up the necessary build steps. Many places are still untyped (i.e. `any`-typed). This will be gradually tackled in the future
* `eslint` is reconfigured accordingly to work for Typescript. Some minor issues were fixed in our code base, but some linter rules, e.g. "don't use `any`", are still disabled as it would require a larger effort to get our code compliant with those rules.
* `prettier` was added to provide code formatting (i.e. consistent whitespace etc.). I mostly stayed with the default config and reformatted the code base accordingly
* Some basic continous integration tests were set-up using GithubActions. You can find an example run under https://github.com/vogelsgesang/query-graphs/runs/791443556?check_suite_focus=true


vogelsgesang:
> The only thing you might do is resolve the "no new line at end of file" warnings in ~5 new files.
Good catch! Fixed

PR #59, opened 2020-08-11
Make tree rendering less verbose

vogelsgesang:
These two commits are both about displaying less information, in particular about hiding irrelevant tooltips and irrelevant statistics about the rendered tree

### DIFF
--- a/query-graphs/src/tree-rendering.ts
+++ b/query-graphs/src/tree-rendering.ts
@@ -765,7 +765,7 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
         nodeUpdate
             .filter(function(d) {
                 return (
-                    Object.getOwnPropertyNames(getDirectProperties(d)).length || // eslint-disable-line indent
+                    Object.getOwnPropertyNames(getDirectProperties(d.data)).length || // eslint-disable-line indent
                     (d.data.hasOwnProperty("properties") && Object.getOwnPropertyNames(d.data.properties).length)
                 );
             }) // eslint-disable-line indent
--- a/query-graphs/src/tree-rendering.ts
+++ b/query-graphs/src/tree-rendering.ts
@@ -976,17 +976,18 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
 
     // Add metrics card
     let treeText = "";
-    const properties = treeData.properties ?? {};
-    treeText += buildPropertyList(properties);
-    treeText += buildPropertyList({nodes: totalNodes});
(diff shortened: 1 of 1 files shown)


query-graphs/src/tree-rendering.ts
```diff
@@ -765,7 +765,7 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
         nodeUpdate
             .filter(function(d) {
                 return (
-                    Object.getOwnPropertyNames(getDirectProperties(d)).length || // eslint-disable-line indent
+                    Object.getOwnPropertyNames(getDirectProperties(d.data)).length || // eslint-disable-line indent
```
rickcole, line 768: I don't recall, does "d" always have a "data" to check for length?

query-graphs/src/tree-rendering.ts
```diff
@@ -976,17 +976,18 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
 
     // Add metrics card
     let treeText = "";
-    const properties = treeData.properties ?? {};
-    treeText += buildPropertyList(properties);
-    treeText += buildPropertyList({nodes: totalNodes});
-    if (crosslinks !== undefined && crosslinks.length) {
-        treeText += buildPropertyList({crosslinks: crosslinks.length});
+    treeText += buildPropertyList(treeData.properties ?? {});
+    if (DEBUG) {
```
rickcole, line 980: Is it a DEBUG situation or some other mode? Perhaps some use cases want more metrics not less. Luis?

query-graphs/src/tree-rendering.ts
vogelsgesang: The `data` member exists on each node, yes.
`d` itself is the node provided by the d3-hierarchy library. It contains all the visual layout information (e.g., x and y positions, width, height), while `d.data` contains the "logical" node (i.e.: property maps etc., symbol names, ...) as returned by our format loaders

query-graphs/src/tree-rendering.ts
vogelsgesang: I agree that some use cases might want more metrics (e.g., the number of operators, an overview of which join types the query contains etc.)

However, I would argue, that the current "node count" and "crosslink count" do not provide those metrics.
E.g., on a Hyper plan, the node count includes a lot expressions, and even subparts which just happen to be separate entities in the tree view and are collapsed by default anyway.

As such, I consider the number of visible nodes mostly as an implementation detail of the query-graphs library and how it decides to render the JSON

PR #63, opened 2020-10-30
Better rendering of optimizer steps

vogelsgesang:
These commits update a bunch of query plans to the latest Hyper versions and then go on to improve the way we render query plans if they contain multiple optimizer steps:

*before*
steps-before

*after*
steps-after


plan-dumper/setup.sql
```diff
@@ -1,2 +1,103 @@
...
+   n_comment varchar(152) not null
+);
+
+CREATE TEMP TABLE region (
+   r_regionkey integer not null,
+   r_name char(25) not null,
+   r_comment varchar(152) not null
+);
+
+-- We also laod some data into the tpv-h schema because our query plans depend on cardinality estimates and we want meaningful numbers there
```
luiseeo, line 93: Typos:
- laod -> load
- tpv-h -> tpc-h

luiseeo (approved):
I found a couple of typos in a comment

plan-dumper/setup.sql
vogelsgesang: thanks for catching those!
Fixed :)

PR #64, opened 2020-11-01
bugfixes for rendering of optimizer steps

vogelsgesang:
**Fix crosllinks for optimizersteps**

With the new rendering of optimizersteps, the crosslinks in query plans
were broken. Note that operatorIds are only unique within an
optimizerstep and not globally

**Don't show empty tooltips**

We don't want to show empty tooltips. Still, we did so due to a bug in
our check if the tooltip would be empty...


rickcole (approved):
Looks good!

PR #65, opened 2020-11-01
Upgrade typescript & code cleanups

vogelsgesang:
This series of commits:
* upgrades the TypeScript compiler and the linter
* updates our callbacks to use the `=>` lambda notation, thereby making them shorter/more concise
* fixes small bugs regarding our animations of cross-links
* removes the usage of `x0`/`y0` from a few places which didn't need them
* simplifies the usage of `x0`/`y0` by using a consistent coordinate system
* moves the symbol definitions into a separate file
and a few other simplifications.

All in all, with this commit we save over 200 lines of code


rickcole (approved):
Great.

PR #66, opened 2020-11-02
Add sort icon

vogelsgesang:
Add an icon for Sort operators

sort


luiseeo (approved):
There's a check failing. But the new icon looks great

PR #67, opened 2020-11-07
Toolbar

luiseeo:
Added toolbar:
- Toolbar icons: 6 new icons.
- Toolbar graphics. The toolbar is included on the top-right card. It displays the 6 icons and has some hover effects (including a tooltip describing what the button does)
- Toolbar functionality. Some old functionality was just mapped to a button, but there is new functionality too.
  - Zoom in
  - Zoom out
  - Rotate left. Most functionality was already present. Rotation basically triggers a refresh.
  - Rotate right
  - Center root.
  - Fit to screen. New! All of the graph can be viewed at once. Centered and zoomed so that everything is visible.
- Corresponding CSS style changes

Also, I saw and fixed a couple of bugs:
- Orientations right-to-left and bottom-to-top had the incorrect initial coordinates (fix was on "orientations" rootx/rooty functions
- Tableau loader streamline collapse was not working correctly. I restored an old function in `tableau.ts` function `collapseNodes`. 

Screen Shot 2020-11-06 at 3 36 25 PM

### DIFF
--- a/query-graphs/src/tree-rendering.ts
+++ b/query-graphs/src/tree-rendering.ts
@@ -1,6 +1,6 @@
 // Import local modules
 import * as treeDescription from "./tree-description";
-import {TreeDescription, Crosslink} from "./tree-description";
+import {TreeDescription, Crosslink, GraphOrientation} from "./tree-description";
 import {defineSymbols} from "./symbols";
 
 // Third-party dependencies
@@ -119,7 +119,7 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
                 return [11.2 /* table node diameter */ + 2, maxLabelLength * 6 + 10 /* textdimensionoffset */];
             },
             nodesep: (a, b) => (a.parent === b.parent ? 1 : 1.5),
-            rootx: scale => -(viewerWidth / 2 - maxLabelLength * 6) / scale,
+            rootx: scale => viewerWidth - (viewerWidth / 2 - maxLabelLength * 6) / scale,
             rooty: _scale => 0,
         },
         "bottom-to-top": {
@@ -132,7 +132,7 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
             nodesize: () => [maxLabelLength * 6, (maxLabelLength * 6) / 2],
             nodesep: (a, b) => (a.parent === b.parent ? 1 : 1),
             rootx: _scale => 0,
-            rooty: scale => -(viewerHeight / 2 - 50) / scale,
+            rooty: scale => viewerHeight - (viewerHeight / 2 - 50) / scale,
         },
         "left-to-right": {
             link: d3shape.linkHorizontal,
@@ -213,6 +213,7 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
 
     // Initialize tooltip
     const tip = d3tip()
+        .attr("id", "tooltip")
         .attr("class", "qg-tooltip")
         .offset([-10, 0])
         .html(d => {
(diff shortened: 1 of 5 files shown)


query-graphs/src/tree-rendering.ts
```diff
@@ -581,11 +582,52 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
...
+        // Solving for `xfactor` we have: xfactor = fullWidth/width
+        const xfactor: number = fullWidth / bounds.width;
+        // Similarly, we would find out that: yfactor = fullHeight/height;
+        const yfactor: number = fullHeight / bounds.height;
+        // Most likely, the X and Y factor are different. We use the minimum of them to avoid cropping
+        let scaleFactor: number = Math.min(xfactor, yfactor);
+        // Add some padding so that the graph is not touching the edges
+        const paddingPercent = 0.9;
+        scaleFactor = scaleFactor * paddingPercent;
+        console.log("Scale factor", scaleFactor);
```
vogelsgesang, line 605: Do we still need the debug output?
Here and other places

query-graphs/src/tree-rendering.ts
```diff
@@ -581,11 +582,52 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
...
+
     // Scale for readability by a fixed amount due to problematic .getBBox() above
     zoomBehavior.scaleBy(baseSvg, 1.5);
 
     orientRoot();
 
+    const infoCard = d3selection
+        .select(target)
+        .append("div")
+        .attr("id", "info-card")
```
vogelsgesang, line 629: In this case the id seems to be completely unused...

query-graphs/src/tree-rendering.ts
```diff
@@ -213,6 +213,7 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
 
     // Initialize tooltip
     const tip = d3tip()
+        .attr("id", "tooltip")
```
vogelsgesang, line 216: whenever possible we should try to avoid using an `id`.
In the JupyterLab plugin, multiple query-graphs will be rendered at the same time and then this Id won't be unique anymore. If we really need to use an id, let's append some timestamp/random number to it and thereby make it unique.

But do we really need an id here? Can't we just use a class instead?

query-graphs/src/tree-rendering.ts
```diff
@@ -594,12 +636,86 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
...
-    d3selection
-        .select(target)
+    infoCard
         .append("div")
         .classed("qg-tree-label", true)
         .html(treeText);
 
+    // Add toolbar
+    function buildToolbarButton(id: string, description: string) {
+        return `<div id='${id}-button' class='qg-toolbar-button'>
```
vogelsgesang, line 646: same issue with IDs. This won't work for the JupyterLab plugin.

I would recommend to instead use something like 
```
const toolbar = infoCard.append("div").classed("qp-toolbar");
function addToolbarButton(symbol: string, description: string, action: ()=>void) {
    toolbar
       .append("div")
       .classed("qg-toolbar-button")
       .on("click", action)
       .html(`<span class="qg-toolbar-icon">
    <svg viewbox='-8 -8 16 16' width='20px' height='20px'>
      <use href='#${id}-symbol' class="qg-collapsed" />
    </svg>
  </span>
<span class="qg-toolbar-tooltip">${description}</span>`);
}
addToolbarButton("zoom-in", "Zoom In", () =>  zoomBehavior.scaleBy(baseSvg.transition().duration(200), zoomFactor));
```

query-graphs/src/tree-rendering.ts
```diff
@@ -594,12 +636,86 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
...
+  <span class="qg-toolbar-icon" id="zoom-in-button">
```
vogelsgesang, line 647: the span shouldn't have an `id`

query-graphs/src/tree-rendering.ts
```diff
@@ -594,12 +636,86 @@ export function drawQueryTree(target: HTMLElement, treeData: TreeDescription) {
...
+        // Positive number: clockwise rotation
+        // Negative number: counter-clockwise rotation
+        const orientationList: GraphOrientation[] = ["top-to-bottom", "left-to-right", "bottom-to-top", "right-to-left"];
+        const index: number = orientationList.indexOf(orientation);
+        let newIndex: number = (index + shift) % 4;
+        newIndex = newIndex < 0 ? newIndex + 4 : newIndex;
+        return orientationList[newIndex];
+    }
+    function clearQueryGraph() {
+        // Removes the QueryGrpah elements. This function is useful if we want to call `drawQueryTree` again
```
vogelsgesang, line 689: QueryGrpah -> QueryGraph

vogelsgesang (commented):
partial review - still have to review the CSS

query-graphs/src/tree-rendering.ts
luiseeo: removing...

query-graphs/src/tree-rendering.ts
luiseeo: removing...

query-graphs/src/tree-rendering.ts
luiseeo: I'm removing it. I was using the ids mostly so I can quickly select the elements in browser `$("#id")`debugging and/or apply styles to them quickly.

query-graphs/src/tree-rendering.ts
luiseeo: I like the idea of passing the action function to avoid the ids

query-graphs/src/tree-rendering.ts
luiseeo: removed

luiseeo:
Addressed initial PR feedback

query-graphs/style/query-graphs.css
```diff
@@ -6,19 +6,77 @@
...
+
+.qg-toolbar {
+  float: right;
+  display: inline-block;
+  padding: 2px;
+}
+
+.qg-toolbar-button {
+  position: relative;
+  display: inline-grid;
```
vogelsgesang, line 34: `inline-grid` is a pretty heavy-weight/overpowered tool for the simple layout we are trying to achieve here. Also, we aren't really using it to setup any grid anyway.

Afaict, an easier way to achieve the same layout would be:
```
display: inline-block;
vertical-align: bottom;
```

vogelsgesang (approved):

PR #70, opened 2020-12-11
Upgrade D3 version and improve typesafety

vogelsgesang:
The main part of this commit is an upgrade of D3 to the latest version. While at it, we also install the type definitions and adjust our code to interact with D3 in a strictly type manner.

Furthermore, there are two smaller "cleanup" commits:
* remove an unused `computeOrder` function
* upgrade the TypeScript compiler


luiseeo (commented):
You got some minor linting issues, but other than that it looks good

Issue #56, opened 2020-08-05
Redesign tooltip; make it possible to display multiple tooltips at the same time


vogelsgesang:
Right now, it is only possible to display detailed information about 1 node at the same time.
This leads to usability issues in a few use cases:

* Comparing two plans: If I open 2 plans side-by-side, I can only display the detailed information for one of them at the same time. This makes it pretty hard to compare the details of the plans with each other
* Compare operators within the same plan: Again, I can only open one tooltip at the same time, so I can't efficiently compare them with each other
* Tooltip obscures parts of the tree: Sometimes (e.g. when comparing cardinality estimates and checking them against information from the tooltip), I have the issue that I can't see other relevant information rendered as part of the normal tree

I am not yet aware of a good solution.
Maybe, we should completely move away from tooltips and rather use some form of "expandable" nodes, similar to how Maya (3d modelling software) allows to expand/collapse nodes in their Node editor...
Not sure how this would influence the graph layouting, though...

rickcole:
Perhaps mark and/or lasso a set of nodes for detailed inspection.
A short term mouse option might be to right click an open tooltip to lock it in place, or similarly a keybind.
Yet another idea, which I've seen before, to put detailed information in a side panel.

vogelsgesang:
> Perhaps mark and/or lasso a set of nodes for detailed inspection.
> A short term mouse option might be to right click an open tooltip to lock it in place, or similarly a keybind.
Both those interactions would allow us to specify which tooltips should stay visible.
However, having multiple tooltips visible at the same time would probably mean that those tooltips overlap each other...

> Yet another idea, which I've seen before, to put detailed information in a side panel.
This is a good solution to solve the "overlaps parts of the graph below the tooltip" issue, however I am not currently sure how it would work for having details for multiple nodes displayed at the same time

vogelsgesang:
In case we went for "expandable/collapsable nodes", we could use https://github.com/klortho/d3-flextree to layout a tree with varying node sizes

vogelsgesang:
I took a look at a few other plan viewers and how they solve the problem

## Pev

https://dalibo.github.io/pev2/#/plan; source code: 
https://tatiyants.com/pev/#/plans/plan_1596705098890

![Screen Shot 2020-08-06 at 11 13 33](https://user-images.githubusercontent.com/6820896/89514394-172cb300-d7d6-11ea-9f8e-3e5ec0ab5357.png)

[...]

PR #72, opened 2020-12-15
Small layout fixes

vogelsgesang:
A bunch of small layout fixes...


query-graphs/src/tree-rendering.ts
```diff
@@ -58,22 +58,22 @@ const orientations: {[k in GraphOrientation]: Orientation} = {
         x: (d, viewSize) => viewSize.y - d.y,
         y: (d, _viewSize) => d.x,
         textdimension: "x",
-        textdimensionoffset: d => (d.children || d._children ? 10 : -10),
-        textanchor: d => (d.children || d._children ? "start" : "end"),
+        textdimensionoffset: d => (d.children ? 10 : -10),
+        textanchor: d => (d.children ? "start" : "end"),
         nodesize: maxLabelLength =>
             [11.2 /* table node diameter */ + 2, maxLabelLength * 6 + 10 /* textdimensionoffset */] as d3point,
```
rickcole, line 64: Could have a minimum here as well? Queries without the Plan and Actual row counts on the labels are fine, but with the row counts there's overlap. It's far from perfect anyway. I suppose the counts might better moved from the middle of the edge. Modified postgres/limitscan.json:

[
  {
    "Plan": {
      "Node Type": "Limit",
      "Parallel Aware": false,
      "Startup Cost": 0.00,
      "Total Cost": 0.01,
      "Plan Rows": 1,
      "Plan Width": 8,
      "Actual Startup Time": 0.001,
      "Actual Total Time": 0.001,
      "Actual Rows": 0,
      "Actual Loops": 1,
      "Output": ["r1", "r2"],
      "Plans": [
        {
          "Node Type": "Result",
          "Parent Relationship": "Outer",
          "Parallel Aware": false,
          "Startup Cost": 0.00,
          "Total Cost": 0.01,
          "Plan Rows": 1,
          "Plan Width": 8,
          "Actual Startup Time": 0.000,
          "Actual Total Time": 0.000,
          "Actual Rows": 0,
          "Actual Loops": 0,
          "Output": ["r1", "r2"]
        }
      ]
    },
    "Planning Time": 0.028,
    "Triggers": [
    ],
    "Execution Time": 0.013
  }
]

query-graphs/src/tree-rendering.ts
```diff
@@ -83,8 +83,8 @@ const orientations: {[k in GraphOrientation]: Orientation} = {
         x: (d, _viewSize) => d.y,
         y: (d, _viewSize) => d.x,
         textdimension: "x",
-        textdimensionoffset: d => (d.children || d._children ? -10 : 10),
-        textanchor: d => (d.children || d._children ? "end" : "start"),
+        textdimensionoffset: d => (d.children ? -10 : 10),
+        textanchor: d => (d.children ? "end" : "start"),
         nodesize: maxLabelLength =>
             [11.2 /* table node diameter */ + 2, maxLabelLength * 6 + 10 /* textdimensionoffset */] as d3point,
```
rickcole, line 89: Same as above.

rickcole (approved):

query-graphs/src/tree-rendering.ts
vogelsgesang: Added a minimum space of 90 pixel.

That solves the issue for your example and seems like a good idea in general.
However, it does not completely solve the issues, e.g. if instead of "Limit" the node label was a longer string. I don't know of a generally nice solution for left-to-right/right-to-left. I guess one possibility would be to move the labels below the nodes, also in the horizontal layouts... Something we might want to do as part of the re-design for #56 anyway...

PR #75, opened 2021-03-08
Add (Postgres) metrics coloring via node rect

rickcole:


query-graphs/src/postgres.ts
```diff
@@ -281,6 +281,70 @@ function colorForeignScan(node: TreeNode, foreignScan?: string) {
...
+
+        const relativeExecutionTime = (nodeTotalTime * nodeLoops - childrenTime) / executionTime;
+        // TODO: remove Actual Total Time of a CTE from referencing CTE Scan subplans
+        // TODO: assert(relativeExecutionTime >= 0, "Unexpected relative execution time");
+
+        assert(node.properties !== undefined);
+        node.properties.set("relativeExecutionTime", relativeExecutionTime.toFixed(3));
+        const l = (100 + (36 - 100) * relativeExecutionTime).toFixed(3);
+        const hsl = "hsl(309, 84%, " + l + "%)";
+        node.properties.set("metricsFill", hsl);
```
vogelsgesang, line 340: Putting this into the `properties` map bypasses the type-safety which we tried to reach by using TypeScript.
Instead rather extend the `TreeNode ` class with a new member.
That way, the supported functionality will be better documented/we have a clearer interface between the loaders and the tree-rendering of QueryGraphs

vogelsgesang (approved):

query-graphs/src/postgres.ts
rickcole: Moved fill attributes into TreeNode.
Added fill attributes to debug properties construction.
Kept relative execution time (not used in tree rendering) in properties for the user.
Renamed relative execution time to collate after existing properties without a resort and to locate consistently.

PR #80, opened 2022-03-01
React-based server-less web app

vogelsgesang:
This series of commits contains a complete rewrite of our "shell" functionality, i.e. the file selection and uploading.

We now have a React-based app which can work completely off-line, and independently of any server.
It can thereby be used directly on https://tableau.github.io/query-graphs/ (currently not yet available; preview version on https://vogelsgesang.github.io/query-graphs/) without reliance on the deployment of our custom `upload-server.js` server.

The landing page of the new app is a rather minimal file picker which also allows copy & paste and drag & drop.
The interaction with the new file opener is more streamlined than our previous landing page. E.g., in general there is no more scrolling necessary, as the complete content now usually fits on a single screen. At the same time, one no longer needs to hit the "submit" button manually, the plan is immediately opened after copy&pasting / dropping the plan onto the file opener.

Furthermore our examples list is now also available on Github, under https://vogelsgesang.github.io/query-graphs/examples.html This page was previously generated ad-hoc by the `upload-server.js`, now it is built as part of the build step for our web-app and then served as a static webpage from Github pages.

To be still able to profit from shareable links, the new frontend  can be optionally integrated with the classical `upload-server,js`: The new `upload-server.js` will redirect its visitors to query-graphs interface hosted on Github Pages, but will set a special `?uploadServer` parameter which will instruct the frontend to upload the plan to the upload server and then construct a shareable URL for that upload.


luiseeo (approved):
Looks great. I appreciate the detailed notes for README and DEVELOPMENT

PR #138, opened 2026-07-15
Support kebab-case Hyper plan JSON keys (backward compatible)

PerFuchs:
Hyper PR hyper-db#13364 renamed the plan JSON property keys from camelCase to kebab-case (e.g. `operatorId` → `operator-id`, `debugName` → `debug-name`, `valueForComparison` → `value-for-comparison`). Only keys changed — enum/tag values (operator tags, join types, window/aggregate function names) keep their camelCase spelling. This PR adds support for the new casing while keeping backward compatibility with the old casing.

The Hyper loader is largely key-agnostic, but it reads three renamed keys by literal name; both spellings are now accepted:

- `operator-id` / `operatorId` (operator id map + crosslink resolution)
- `debug-name` / `debugName` (display name + always-shown property)
- `value-for-comparison` / `valueForComparison` (fixed child ordering)

Example plans are regenerated in the new kebab-case format, matching what `dump-plans.py` now emits against a Hyper build containing hyper-db#13364, so the default examples reflect current output. Keys are renamed using the exact mapping from hyper-db#13364 (137 keys, including acronym cases such as `resultIUs` → `result-ius` and `filterJoinIU` → `filter-join-iu`); keys that #13364 left untouched (e.g. `distinctValues`, `allowTrustedCasts`) are kept as-is. `tablescan-legacy-camelcase.plan.json` is kept as the single old-syntax example to exercise the backward-compatibility path.


standalone-app/examples/hyper/tablescan-kebab-case.plan.json
vogelsgesang, line 1: we usually regenerate all query plans by rerunning the dump_plan script, with locally overriding HyperAPI

Maybe add one test for the old syntax instead

vogelsgesang (approved):

standalone-app/examples/hyper/tablescan-kebab-case.plan.json
PerFuchs: Good call — done. I regenerated all Hyper example plans in the new kebab-case format (using the exact key mapping from hyper-db#13364, including acronym cases like resultIUs → result-ius), removed the redundant tablescan-kebab-case.plan.json, and kept a single tablescan-legacy-camelcase.plan.json to cover the old syntax. I verified every example still parses through loadHyperPlanFromText, and that the legacy debugName / new debug-name both resolve the display name correctly.

PR #139, opened 2026-08-18
Visualize execution pipelines from EXPLAIN (FORMAT JSON, PIPELINES)

PerFuchs:
## What this adds

Visualizes Hyper's execution **pipelines** from `EXPLAIN (FORMAT JSON, PIPELINES)`
(hyper-db-emu/hyper-db#13438) on top of the operator tree — using **colour only**.
Plans without pipeline data render exactly as before.

## The input

With `PIPELINES`, the plan is wrapped in an envelope:

```json
{
  "tree":  { ...operator tree, each node has an "operator-id"... },
  "pipelines": [ { "id": 4, "operators": [3,4,5,6], "dependencies": [] }, ... ]
}
```

`pipelines[].operators` is an unordered set of tree operator ids; one operator can
belong to several pipelines (e.g. a `UNION ALL` target nested into each input).

## How it looks (colour only)

- Each operator carries a small colour **bar above** it (the pipelines it shares
  with its **parent** — its outgoing side) and a **bar below** it (the pipelines
  it shares with its **children** — its incoming side), split into one segment
  per pipeline.
- The operator **icon** is tinted with the node's right-most pipeline colour
  (this also colours the minimap).
- The **incoming edge** is stroked with the pipeline(s) that flow across it: a
  solid colour for one, a contiguous **gradient** for several (e.g. the edge
  above a `UNION ALL` target). An edge whose endpoints share no pipeline stays
  neutral grey, so **pipeline breakers** are visible.

## Colouring rules

- A pipeline is coloured the **first time it is encountered** in a single
  pre-order DFS, taking the next palette colour (a 20-colour Tableau palette,
  rotating once exhausted). Because colours follow **tree position, not pipeline
  ids**, renumbering the pipelines between executions does not change any colour.
- Bar segments are ordered **left-to-right** by the child branch that carries
  each pipeline, so they line up with the tree.
- Shared operators read naturally: the right-most pipeline wins for the icon;
  a fork/share source keeps **one consistent colour** under both arms; a
  `UNION ALL` target shows all of its arms' colours.

## Where to look

- `query-graphs/src/hyper.ts` — detect the `{tree, pipelines}` envelope and the
  single-pass `assignPipelineColors` (first-appearance colouring + bar/edge/icon
  assignment in one DFS).
- `query-graphs/src/tree-description.ts` — the presentation model stays
  colour-only: `barsAbove`, `barsBelow`, `edgeColors` (arrays of colours; the
  "pipeline" concept never leaks into it).
- `query-graphs/src/ui/QueryNode.tsx` (+ `.css`), `PipelineEdge.tsx`,
  `tree-layout.ts`, `QueryGraph.tsx` — rendering.
- `standalone-app/examples/hyper/pipelines-*.plan.json` — example plans
  (fork/share, UNION ALL, magic/EXISTS, group-by, multi-join, nested), taken
  from #13438's fixture where available.

## Screenshots

Real `EXPLAIN (FORMAT JSON, PIPELINES)` plans. Each operator has a colour bar above (pipelines shared with its parent) and below (shared with its children); the icon takes the right-most pipeline colour; edges carry the shared pipeline colour (a gradient when several, grey for a pipeline breaker).

**Fork / share** — the shared fork source stays one consistent colour under both `groupBy` arms; the join's two inputs are distinct.

fork/share

**Materialized share** — the same query with share forking disabled: the share becomes its own pipeline and both `groupBy` arms read it across a pipeline break (grey edges), instead of the shared scan recurring across pipelines.

materialized share

**UNION ALL** — the three arms are distinct, and the union target carries all three (multi-segment bars + a gradient edge above it).

union all

**Magic unnesting** — the existing `magicunnesting` example dumped with `PIPELINES`: a correlated scalar subquery lowered to a magic set, with shared pipelines across the magic `groupBy` and the outer join.

magic / EXISTS

**Multi-join (5 pipelines)** — the hot spine and the fact/dimension scans read as distinct pipelines.

joins

**Hovering** a node highlights it; the colour bars stay attached above and below.

hovering a node

**Expanding** a node shows its properties inside the box; the lower bar sits after the properties.

expanded node


PerFuchs:
Note to self: remove the Screenshots section from the PR description before merging — the images are hosted on a temporary `pr-media` branch of my fork (referenced via raw URLs), so they shouldn't remain in the permanent merge commit message.

query-graphs/src/tree-description.ts
```diff
@@ -31,6 +31,18 @@ export interface TreeNode {
...
+    // (conceptually the "outgoing" side, toward the parent).
+    // Empty/undefined means no bar. Segments are drawn in array order.
+    barsAbove?: string[];
+    // Colors of a bar drawn just below the node
+    // (conceptually the "incoming" side, toward the children).
+    // Empty/undefined means no bar. Segments are drawn in array order.
+    barsBelow?: string[];
+    // Colors of the incoming edge. Several colors are drawn as a gradient,
+    // a single color as a solid stroke.
+    edgeColors?: string[];
```
vogelsgesang, line 44: Group with existing properties

```suggestion
    // Colors of a bar drawn just above the node
    // (conceptually the "outgoing" side, toward the parent).
    // Empty/undefined means no bar. Segments are drawn in array order.
    barsAbove?: string[];
    // Colors of a bar drawn just below the node
    // (conceptually the "incoming" side, toward the children).
    // Empty/undefined means no bar. Segments are drawn in array order.
    barsBelow?: string[];

    // Additional CSS classes applied to the incoming link
    edgeClass?: string;
    // Label placed on the incoming edge
    edgeLabel?: string;
    // Width of the incoming edge
    edgeWidth?: number;
    // Colors of the incoming edge. Several colors are drawn as a gradient,
    // a single color as a solid stroke.
    edgeColors?: string[];
```

standalone-app/examples/hyper/nested-except-over-unionall.plan.json
vogelsgesang, line 1: Why do we have new test cases? Why without the corresponding SQL?
How would we ever re-generate those in the future?

Why did we not update the existing query plans?

query-graphs/src/ui/tree-layout.ts
```diff
@@ -80,9 +80,11 @@ export function layoutTree(
             id: `${sourceId}->${targetId}`,
             source: sourceId,
             target: targetId,
+            type: "pipeline",
```
vogelsgesang: the "UI" layer should be database-indepedent. Don't call it "pipeline" on this level

query-graphs/src/ui/PipelineEdge.tsx
```diff
@@ -0,0 +1,69 @@
...
+    colors?: string[];
+}
+
+// A tree edge that can carry multiple colors.
+//
+// An edge can belong to several pipelines at once (e.g. the edge above a
+// UNION ALL target, executed once per input). When several colors are given the
+// stroke is painted with a gradient of contiguous color bands (one per color,
+// running source->target); a single color is drawn as a solid stroke.
+export function PipelineEdge(props: EdgeProps<PipelineEdgeData>) {
```
vogelsgesang: don't call it "pipeline"; UI is database-independent while "pipeline" is a Hyper-specific detail.

Call it, e.g., `ColoredEdge`

query-graphs/src/ui/PipelineEdge.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const {id, sourceX, sourceY, sourcePosition, targetX, targetY, targetPosition, markerEnd, label, labelStyle, style} = props;
+    const [edgePath, labelX, labelY] = getBezierPath({
+        sourceX,
+        sourceY,
+        sourcePosition,
+        targetX,
+        targetY,
+        targetPosition,
+    });
```
vogelsgesang: Do we really need to repackage the props?

```suggestion
    const [edgePath, labelX, labelY] = getBezierPath(props);
```

query-graphs/src/hyper.ts
```diff
@@ -356,9 +386,131 @@ function convertHyperPlan(node: Json): TreeDescription {
     colorRelativeExecutionTime(conversionState);
     setEdgeWidths(conversionState);
     const crosslinks = resolveCrosslinks(conversionState);
+    if (pipelines !== undefined) {
+        assignPipelineColors(root, conversionState.operatorsById, parsePipelines(pipelines));
+    }
     return {root, crosslinks, metadata: conversionState.metadata};
 }
 
+// A raw pipeline entry, as parsed from the `pipelines` array of the plan.
```
vogelsgesang, line 368: move utility / helper functions before their usage in `convertHyperPlan`

query-graphs/src/hyper.ts
```diff
@@ -356,9 +386,131 @@ function convertHyperPlan(node: Json): TreeDescription {
...
+    for (const p of resolved) {
+        for (const node of p.nodes) {
+            const list = nodePipelines.get(node) ?? [];
+            list.push(p);
+            nodePipelines.set(node, list);
+        }
+    }
+
+    let nextColor = 0;
+    const walk = (node: TreeNode, parent: TreeNode | undefined) => {
```
vogelsgesang, line 435: why do we need to walk the tree?

Can't we simply walk the `resolved` pipelines?
I.e. instead of building the `nodePipelines` map, directly assign colors and bars in the `for` loop ahead?

query-graphs/src/hyper.ts
```diff
@@ -356,9 +386,131 @@ function convertHyperPlan(node: Json): TreeDescription {
...
+// order, not the pipeline ids, so renumbering the pipelines between executions
+// does not change any color.
+//
+// Data flows leaves->root, so a node's children feed it from below (its
+// "incoming" pipelines, drawn as the bar below) and the node feeds its parent
+// above (its "outgoing" pipelines, drawn as the bar above and the incoming
+// edge). A pipeline that starts or ends at a node (a pipeline breaker, or a
+// source/sink) shows on only one side; an edge with no shared pipeline stays
+// neutral. The icon takes the node's right-most pipeline color.
+function assignPipelineColors(root: TreeNode, operatorsById: Map<string, TreeNode>, pipelines: RawPipeline[]): void {
```
vogelsgesang: I think we should have a bar below the leaves and bars above the "output"

query-graphs/src/hyper.ts
```diff
@@ -22,9 +22,43 @@ The main steps are:
 
 */
 
-import {TreeNode, TreeDescription, Crosslink, IconName} from "./tree-description";
+import {TreeNode, TreeDescription, Crosslink, IconName, allChildren} from "./tree-description";
 import {Json, JsonObject, forceToString, tryToString, formatMetric, hasOwnProperty, tryGetPropertyPath} from "./loader-utils";
```
vogelsgesang, line 27: how do explicit scans render? How does fork vs materialize look?

vogelsgesang, line 27: the operator icons in your screenshots are rendered incorrectly. E.g., the `scan` doesn't use the correct icons

standalone-app/examples/hyper/pipelines-analyze-forkshare.plan.json
```diff
@@ -0,0 +1,264 @@
...
+        7
+      ],
+      "dependencies": [
+        1,
+        4
+      ],
+      "statistics": {
+        "cpu-cycles": 12052,
+        "wall-time": 4444
+      }
```
vogelsgesang: where do we want to render those stats in the mid-term?

query-graphs/src/hyper.ts
PerFuchs: We need the tree walk to assign the colors in the bars in the correct order. For example, the bars below a union all match the order of the incoming pipelines. Do you have an alternative in mind?

With regards to assigning colours upfront. That's something we could do but then they would change if the ids change, I rather have them depend on the topology of the tree. This also makes it unlikely to end up with two similar/same colors close to each other.

query-graphs/src/hyper.ts
PerFuchs: That's already the behavior — and I reworded the comment that made it read ambiguously. The root `output` (no parent) always gets a full bar above (every pipeline it drives), and a leaf (a source) always gets a full bar below; only in-between pipeline breakers are one-sided.

query-graphs/src/hyper.ts
PerFuchs: The cause: #13438's FORMAT JSON emits camelCase operator names (`scan`, `groupBy`, `explicitScan`, `output`, …) while the visualizer's icon config keys are the internal lowercase names (`tablescan`, `groupby`, `executiontarget`), so the mismatched ones fall back to the default circle. Since neither spelling is kebab-case (unlike the JSON keys, e.g. `operator-id`), the proper fix is Hyper-side: make the FORMAT JSON operator names kebab-case too, then update the icon config to match. I'll do that as a follow-up once the operator names are fixed, rather than hard-coding camelCase aliases here.

standalone-app/examples/hyper/pipelines-analyze-forkshare.plan.json
PerFuchs: Today we render cpu-cycles as node heat (operators >=5% of total), actual/estimated row counts as the edge label + thickness, and the rest of the `statistics`/`analyze` fields in the node tooltip. For per-pipeline stats: how about a hover on each pipeline edge?

standalone-app/examples/hyper/nested-except-over-unionall.plan.json
PerFuchs: Good push on reproducibility — addressed:

- The three pipeline examples (fork/share, UNION ALL, magic/EXISTS) are the exact expected `FORMAT JSON` output of Hyper's own `hyper/sql/functional/features/explain/pipeline_graph.test` (hyper-db#13438) — I verified them byte-for-byte. I added their source SQL under `plan-dumper/queries/pipelines-*.sql` and taught `dump-plans.py` a `PIPELINES` mode, so they regenerate the same way as every other example: `python3 dump-plans.py --hyper-path <#13438 build>`. The pip-bundled Hyper doesn't support the option yet, which I noted in `DEVELOPMENT.md`.
- The existing plans weren't left stale — they were refreshed to the new format in an earlier commit on this branch.
- I dropped the cases that had no reproducible source: this `nested-except-over-unionall` one, and the ANALYZE pipeline plans (their `cpu-cycles` are non-deterministic, so they can't be stable golden files).

PerFuchs: sql + new dumping mode have been added

query-graphs/src/hyper.ts
PerFuchs: see screenshots in the PR description. I added the materialized one today. Let me know if you meant something different.

DEVELOPMENT.md
```diff
@@ -73,6 +73,9 @@ python3 dump-plans.py --hyper-path ~/workspace/hyper-db/bazel-bin/hyper/tools/hy
 The `--hyper-path` argument should point to the directory containing the `hyperd` binary.
 When omitted, the script uses the `hyperd` bundled with the pip-installed `tableauhyperapi` package.
 
+The pipeline examples (`queries/*-pipelines.sql`) use the `EXPLAIN (… PIPELINES …)` option
+and are only dumped for Hyper (skipped for Postgres).
+
```
vogelsgesang: doesn't belong here. Other modes are also not documented here.

If you want to document this, document all modes in `plan-dumper/README.md`

query-graphs/src/hyper.ts
vogelsgesang: sorry, I missed the *not*....

I think we should *not* have a bar below the leaves and bars above the "output"

was adding the bars intentional or just an accident? Intuitively I would have expected them to not be there

query-graphs/src/hyper.ts
vogelsgesang: ok, so they are broken :/

plan-dumper/dump-plans.py
```diff
@@ -37,15 +38,40 @@ def dump_plans(name, exec_stmt, get_plan):
...
+       elif fname.endswith("-analyze.sql"):
+           mode = "analyze"
        else:
-           plan = get_plan(sql, None)
+           mode = None
+
+       # Pipeline query files are self-contained scripts: setup (SET/CREATE), the
+       # query to EXPLAIN, then teardown that resets the session globals and drops
+       # the temp tables. Only Hyper supports the PIPELINES option. Run the setup
+       # now and the teardown after, so only the query itself is EXPLAINed.
```
vogelsgesang: If we introduce such a mechanism, make it a general one.

That being said, I am not convinced, yet, that we actually need it.
Better test queries shouldn't need multiple statements

plan-dumper/dump-plans.py
```diff
@@ -98,6 +124,10 @@ def get_hyper_plan(sql, mode):
                 explain = "EXPLAIN (FORMAT JSON, OPTIMIZE STEPS) "
             elif mode == "analyze":
                 explain = "EXPLAIN (FORMAT JSON, ANALYZE) "
+            elif mode == "pipelines":
+                explain = "EXPLAIN (FORMAT JSON, PIPELINES, EXPAND_VIEWS true, EXPRESSIONS SQL) "
+            elif mode == "analyze-pipelines":
+                explain = "EXPLAIN (FORMAT JSON, PIPELINES, ANALYZE, EXPAND_VIEWS true, EXPRESSIONS SQL) "
             elif mode is None:
                 explain = "EXPLAIN (FORMAT JSON) "
```
vogelsgesang: Are we actually intentionally testing the external instead of the internal representation?

I think we should have tests for both. We should add `JSON` as soon as it's ready. In the meantime, we should probably still stick to `INTERNAL_JSON`?

```suggestion
                explain = "EXPLAIN (FORMAT INTERNAL_JSON, OPTIMIZE STEPS) "
elif mode == "analyze":
                explain = "EXPLAIN (FORMAT INTERNAL_JSON, ANALYZE) "
            elif mode == "pipelines":
                explain = "EXPLAIN (FORMAT INTERNAL_JSON, PIPELINES, EXPAND_VIEWS true, EXPRESSIONS SQL) "
            elif mode == "analyze-pipelines":
                explain = "EXPLAIN (FORMAT INTERNAL_JSON, PIPELINES, ANALYZE, EXPAND_VIEWS true, EXPRESSIONS SQL) "
            elif mode is None:
                explain = "EXPLAIN (FORMAT INTERNAL_JSON) "
```

query-graphs/src/ui/ColoredEdge.tsx
vogelsgesang, line 1: how do the bars above / below look when hovering or expanding a node?

standalone-app/examples/hyper/nested-except-over-unionall.plan.json
PerFuchs: Reworked: the pipeline examples are now plain single-statement SQL over the shared `t1`/`t2` tables and dumped exactly like every other example (no bespoke setup). This nested ad-hoc case was dropped.

query-graphs/src/hyper.ts
PerFuchs: It was intentional, but you're right it reads better without — removed. A bar is now only drawn where a pipeline actually crosses a tree edge, so leaves have no bar below and the output none above.

query-graphs/src/hyper.ts
PerFuchs: Fixed. The icons fell back to the default circle because the examples used the external FORMAT JSON operator names; the dumper now uses FORMAT INTERNAL, whose operator names match the icon config. That also fixed the missing edges — the explicit-scan → share crosslinks weren't rendering before either. I also added a materialized-share example, so fork vs materialize reads clearly now. The screenshots in the description are updated.

PerFuchs: Resolved without aliases: the dumper now emits FORMAT INTERNAL, whose operator names already match the icon config, so the icons render correctly.

DEVELOPMENT.md
PerFuchs: Agreed — removed it (the other modes aren't documented here either).

plan-dumper/dump-plans.py
PerFuchs: Good call — dropped the mechanism entirely. The pipeline examples are now single-statement queries: a selective filter keeps the shared scan from being inlined, and fork-vs-materialize falls out of the query shape (independent consumers fork the share, a self-join materializes it), so no setup statements are needed.

plan-dumper/dump-plans.py
PerFuchs: Done — the dumper now uses FORMAT INTERNAL for all modes. Agreed we can add external FORMAT JSON coverage once it's ready.

query-graphs/src/ui/ColoredEdge.tsx
PerFuchs: I added screenshots, they appear around the symbol. No strong opinion here. We can also have them outside of the boxes. Let me know what you prefer.

plan-dumper/queries/magic-pipelines.sql
vogelsgesang, line 1: how is this different from [plan-dumper/queries/magicunnesting.sql](https://github.com/tableau/query-graphs/blob/main/plan-dumper/queries/magicunnesting.sql)?

I prefer that you added a comment; but maybe we should now remove `magicunnesting.sql`?

query-graphs/src/ui/ColoredEdge.tsx
vogelsgesang: I like them inside the boxes :)

But for

expanded node

I wonder if the lower bar should be after the properties instead of between the icon and the props

standalone-app/examples/hyper/pipelines-analyze-forkshare.plan.json
vogelsgesang: Not sure about a hover...

Problem with hovers is that you can't have multiple of them open at the same time and comparing stats between two plans is hard since you can not hover pipelines in both of them at the same time

I wonder if we will have a side-bar for pipeline metrics eventually...

query-graphs/src/ui/ColoredEdge.tsx
PerFuchs: Moved the lower bar so it sits after the properties when the node is expanded (and still under the icon when collapsed). Screenshot in the description is updated.

plan-dumper/queries/magic-pipelines.sql
PerFuchs: Yes — same magic-set plan shape, just a different query (EXISTS vs the existing scalar subquery). Switched the PIPELINES example over to `magicunnesting.sql` and dropped the extra EXISTS variant. Screenshot is from that plan now.

vogelsgesang (approved):

PR #141, opened 2026-08-21
Support current FORMAT INTERNAL plans in Hyper visualizer

PerFuchs:
Accepts the kebab-case operator and expression tags introduced by Hyper PR #13883 while preserving legacy separatorless tags. It also accepts the unified `scan`, `join`, and `insert` operators and their `type` fields introduced by Hyper PR #13918 while preserving older operator-specific tags, using subtype-specific join icons and crosslinks and retaining the virtual-table scan icon. Current `statistics`, `estimated-rows`, and `output-rows` fields are supported alongside their older spellings.


plan-dumper/dump-plans.py
```diff
@@ -95,11 +95,11 @@ def exec_hyper(sql):
...
         def get_hyper_plan(sql, mode):
             if mode == "steps":
-                explain = "EXPLAIN (FORMAT JSON, OPTIMIZE STEPS) "
+                explain = "EXPLAIN (FORMAT INTERNAL, OPTIMIZE STEPS) "
             elif mode == "analyze":
-                explain = "EXPLAIN (FORMAT JSON, ANALYZE) "
+                explain = "EXPLAIN (FORMAT INTERNAL, ANALYZE) "
             elif mode is None:
-                explain = "EXPLAIN (FORMAT JSON) "
+                explain = "EXPLAIN (FORMAT INTERNAL) "
```
vogelsgesang, line 162: I think we want to test both, internal and external modes...
Not 100% sure how to do this, though

PerFuchs: Agreed, but I would not want to do this in this PR.

query-graphs/src/loaders/hyper.ts
```diff
@@ -82,43 +82,86 @@ interface NodeRenderingConfig {
 }
 
 const nodeRenderingConfig: Record<string, NodeRenderingConfig> = {
```
vogelsgesang, line 84: don't duplicate all those entries.

Rather add a separate "legacy names" mapping which maps the old to the new keys. That way, we at least have all the "legacy" stuff centralized in a single place

query-graphs/src/loaders/hyper.ts
```diff
@@ -177,7 +225,7 @@ function convertHyperNode(rawNode: Json, parentKey, conversionState: ConversionS
         // Display these properties always as properties, even if they are more complex.
         // `debugName` is the pre-kebab-case spelling of `debug-name`; we accept both for
         // backwards compatibility with plans produced before the Hyper kebab-case cutover.
-        const propertyKeys = ["debug-name", "debugName", "analyze", "sqlpos"];
+        const propertyKeys = ["debug-name", "debugName", "statistics", "analyze", "sqlpos"];
```
vogelsgesang, line 228: are the old keys already deprecated for a while? If so, let's remove them

```suggestion
        const propertyKeys = ["debug-name", "statistics", "sqlpos"];
```

query-graphs/src/loaders/hyper.ts
```diff
@@ -265,21 +313,31 @@ function convertHyperNode(rawNode: Json, parentKey, conversionState: ConversionS
...
+            tryGetPropertyPath(rawNode, ["statistics", "running"]) ?? tryGetPropertyPath(rawNode, ["analyze", "running"]);
+        const errored = conversionState.metadata.has("Error") && running === true;
         if (errored) {
             convertedNode.iconColor = "red";
         }
 
         // Information on the execution time
-        const execTime = tryGetPropertyPath(rawNode, ["analyze", "cpu-cycles"]);
+        const execTime =
+            tryGetPropertyPath(rawNode, ["statistics", "cpu-cycles"]) ?? tryGetPropertyPath(rawNode, ["analyze", "cpu-cycles"]);
```
vogelsgesang, line 325: remove support for the old one; here and other places

query-graphs/src/loaders/hyper.ts
```diff
@@ -265,21 +313,31 @@ function convertHyperNode(rawNode: Json, parentKey, conversionState: ConversionS
         } as TreeNode;
 
         // Highlight the node which errored out, in case the query failed
-        const errored = conversionState.metadata.has("Error") && tryGetPropertyPath(rawNode, ["analyze", "running"]) === true;
+        const running =
+            tryGetPropertyPath(rawNode, ["statistics", "running"]) ?? tryGetPropertyPath(rawNode, ["analyze", "running"]);
```
vogelsgesang, line 317: why does the `running` flag even still exist on the operator? Should have been moved into the `pipeline`?

query-graphs/src/loaders/hyper.ts
```diff
@@ -265,21 +313,31 @@ function convertHyperNode(rawNode: Json, parentKey, conversionState: ConversionS
...
+        const estimatedCard =
+            hasOwnProperty(rawNode, "estimated-rows") && typeof rawNode["estimated-rows"] === "number"
+                ? rawNode["estimated-rows"]
+                : hasOwnProperty(rawNode, "cardinality") && typeof rawNode.cardinality === "number"
+                  ? rawNode.cardinality
+                  : undefined;
+        if (estimatedCard !== undefined) {
+            const actualCard =
+                tryGetPropertyPath(rawNode, ["statistics", "output-rows"]) ??
+                tryGetPropertyPath(rawNode, ["analyze", "tuple-count"]);
```
vogelsgesang, line 340: simplify; only support the new properties

Here and all other places

plan-dumper/dump-plans.py
vogelsgesang: Why not?

Note that I introduced a new `VARIANTS` mechanism in the plan dumper which should make this much more straightforward... Just add the external JSON for a few interesting queries, and double-check that they work. Afaik, they should work out of the box?

vogelsgesang:
superseded by #167

PR #158, opened 2026-08-24
Tighten pipeline colour-bar spacing to outgoing edges

PerFuchs:
## Summary
- The 1em margin under every below-bar (added so the subtree +/- would not sit on the bar) left a large gap between operators and the edges below them.
- Drop that margin and sit the subtree handle slightly lower (`bottom: -1.15em`) so the + still clears the bar without stretching the node.

## Test plan
- [ ] Open `examples.html` forkshare / materialized / unionall / magicunnesting pipeline plans
- [ ] Edges should sit close under the lower colour bars
- [ ] Hover a node with collapsed children: the +/- handle should not overlap the bar
- [ ] Expand a node: the lower bar should still sit after the properties

### DIFF
--- a/query-graphs/src/ui/QueryNode.css
+++ b/query-graphs/src/ui/QueryNode.css
@@ -23,7 +23,8 @@
     font-size: .7em;
     width: 1em;
     height: 1em;
-    bottom: -.8em;
+    /* Sit below the colour bar instead of overlapping it. */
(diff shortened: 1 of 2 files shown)


PerFuchs:
Closed as superseded by this PR: https://github.com/tableau/query-graphs/pull/169

PR #170, opened 2026-09-04
Prevent query graph flickering during relayout and fix minimap

vogelsgesang:
So far, there was some flickering of the edge labels. This was caused because we recreated the node objects we passed into reactflow and reset the `measured` dimensions. Therefore, reactflow re-measured the nodes and while doing so hid the labels.

We now preserve React Flow's measured dimensions across layout updates so existing nodes remain initialized,
This also fixes the minimap.

We use React Flow's dimension changes as the sole measurement source and remove the parallel ResizeObserver.

Drive-by fix: each QueryGraph now owns its rendering store, decoupling expansion, subtree, and measurement state between multiple graph instances on the same page.


query-graphs/src/ui/QueryGraph.tsx
```diff
@@ -31,59 +36,27 @@ const edgeTypes = {
...
-            allChildren,
-        );
-        initGraphStore(expandedSubtrees);
-    }, [treeDescription, initGraphStore, nodeIdMapping]);
-
-    // Create a ResizeObserver to keep track of the sizes of the nodes
-    const resizeObserverRef = useRef<ResizeObserver | undefined>(undefined);
+function QueryGraphInternal({treeDescription, children, nodeIdMapping}: QueryGraphInternalProps) {
+    // Keep React Flow's measurements in the controlled node objects. Dropping them when
+    // recomputing the layout makes React Flow repeatedly hide and re-initialize the nodes.
```
vogelsgesang, line 41:
```suggestion
    // Keep React Flow's measurements in the controlled node objects. Dropping them when
    // recomputing the layout would cause React Flow to re-initialize the nodes, leading to visible
    // flickering of the edge labels.
```

query-graphs/src/ui/QueryGraph.tsx
```diff
@@ -31,59 +36,27 @@ const edgeTypes = {
...
+            });
+            updateNodeDimensions(updates);
+        },
+        [updateNodeDimensions],
+    );
 
-    // Layout the tree, using the actual measured sizes of the DOM nodes
-    const nodeDimensions = useGraphRenderingStore((s) => s.nodeDimensions);
-    const expandedNodes = useGraphRenderingStore((s) => s.expandedNodes);
+    // Layout the tree using the dimensions measured by React Flow itself.
```
vogelsgesang, line 55:
```suggestion
    // Layout the tree using the dimensions measured by React Flow
```

query-graphs/README.md
```diff
@@ -48,10 +48,10 @@ The library intentionally exposes low-level loaders (`json`, `xml`) as generic f
 ## The Renderer
 
 `QueryGraph` (`src/ui/QueryGraph.tsx`) is the top-level component rendering a `TreeDescription`.
-It assigns a stable id to every node, seeds the interaction store from each node's `expandedByDefault` flag, and observes on-screen node sizes with a `ResizeObserver`.
+It assigns a stable id to every node, creates a graph-local rendering store seeded from each node's `expandedByDefault` flag, and retains the node dimensions measured by react-flow.
 
 `tree-layout.ts` positions the tree with [`d3-flextree`](https://github.com/Klortho/d3-flextree) on top of `d3-hierarchy`, then translates the result into react-flow nodes and edges.
-Layout is driven by the **measured** DOM size of each node, so it runs in two passes: the first render uses a placeholder size and, once the `ResizeObserver` reports real dimensions, the tree re-lays-out with the correct sizes.
+Layout is driven by the **measured** DOM size of each node, so it runs in two passes: react-flow measures new nodes after their first render, then the tree re-lays-out with the correct sizes. Those measurements are retained in the controlled node objects so react-flow does not re-initialize them on every layout.
```
vogelsgesang, line 54:
```suggestion
Layout is driven by the **measured** DOM size of each node, so it runs in two passes: react-flow measures new nodes after their first render, then the tree re-lays-out with the correct sizes.
Those measurements are retained in the controlled node objects so react-flow does not re-initialize them on every layout.
```

query-graphs/src/ui/QueryGraph.tsx
vogelsgesang: 🤖 Addressed

query-graphs/src/ui/QueryGraph.tsx
vogelsgesang: 🤖 Addressed

query-graphs/README.md
vogelsgesang: 🤖 Addressed

PR #171, opened 2026-09-04
Animate query graph layout transitions

vogelsgesang:
Compute the final layout once and interpolate node size and layout from a shared clock, avoiding intermediate relayout jumps.

Animate subtree additions and removals from or toward their parent while fading nodes, edges, and labels. Delay entry animations until React Flow has measured the new nodes and preserve exit destinations across interrupted transitions.

Also, disable pointer interaction for transient elements, to avoid hover animations being influenced by those transient elements.


query-graphs/src/ui/animation.ts
vogelsgesang: does it make sense to move this into `useAnimatedGraphLayout`?

query-graphs/src/ui/animated-layout.ts
```diff
@@ -0,0 +1,147 @@
...
+    opacity: number;
+    transient: boolean;
+    exitPosition?: Position;
+}
+
+interface AnimatedEdge {
+    edge: GraphEdge;
+    opacity: number;
+    transient: boolean;
+}
```
vogelsgesang: document the meaning of "transient"

query-graphs/src/ui/animated-layout.ts
```diff
@@ -0,0 +1,155 @@
...
+    position: Position;
+    opacity: number;
+    transient: boolean;
+    exitPosition?: Position;
+}
+
+interface AnimatedEdge {
+    edge: GraphEdge;
+    opacity: number;
+    transient: boolean;
```
vogelsgesang: document the meaning of "transient"

query-graphs/src/ui/animated-layout.ts
```diff
@@ -0,0 +1,142 @@
...
+    anchors: TransitionAnchors,
+    progress: number,
+): AnimatedLayout {
+    // Union both layouts: new elements emerge from the anchor, while removed
+    // elements remain mounted until they reach the anchor and become transparent.
+    const fromNodes = new Map(from.nodes.map((entry) => [entry.node.id, entry]));
+    const toNodes = new Map(to.nodes.map((node) => [node.id, node]));
+    const nodes: AnimatedNode[] = [];
+
+    for (const id of new Set([...fromNodes.keys(), ...toNodes.keys()])) {
```
vogelsgesang: are we recreating those Maps / Sets on each frame? Should we instead cache them somwhere?

query-graphs/src/ui/animated-layout.ts
```diff
@@ -0,0 +1,142 @@
...
+        const start = fromNodes.get(id);
+        const target = toNodes.get(id);
+        if (target === undefined && progress === 1) continue;
+        const anchor = anchors.get(id);
+        const fromPosition = start?.position ?? anchorPosition(fromNodes, anchor) ?? {x: 0, y: 0};
+        const exitPosition =
+            target === undefined ? (start?.exitPosition ?? anchorPosition(toNodes, anchor) ?? {x: 0, y: 0}) : undefined;
```
vogelsgesang: why do we have both `anchorPosition` and the `exitPosition` property on nodes? can we standardize on a single mechanism?

query-graphs/src/ui/animated-layout.ts
vogelsgesang: 🤖 Yes. `interpolateLayout` rebuilds the source/target indexes and ID union on every frame, and `refreshLayoutData` similarly rebuilds its latest-data indexes. I recommend preparing an interpolator once per transition so those indexes are cached while only positions and opacity are recomputed per frame. That removes repeated O(n) setup without introducing persistent cache invalidation. Does that direction sound good?

query-graphs/src/ui/animated-layout.ts
vogelsgesang: 🤖 `anchorPosition` resolves a node-relative handle for a transition, while `exitPosition` freezes that resolved destination if the transition is retargeted. With the new retarget-from-current behavior, I think the frozen destination is unnecessary and slightly inconsistent: an exiting node should follow the anchor in the latest target. I recommend removing `exitPosition` and using the anchor mechanism exclusively. Does that match your expectation?

query-graphs/src/ui/animated-layout.ts
vogelsgesang: would it make sense to fold `sameTargetLayout` into it, such that the interpolator's factory function simply returns `undefined` if there is nothing to interpolate?

query-graphs/src/ui/animated-layout.ts
vogelsgesang: do it

query-graphs/src/ui/animated-layout.ts
vogelsgesang: 🤖 I would keep that check separate. The interpolator's source is the current rendered state, which intentionally differs from the target throughout an active animation. Comparing those would restart interpolation on payload-only updates. Passing both the previous target and current rendered state into the factory would merely move `sameLayoutTarget` while making the API less focused. I propose caching indexes in the factory while retaining the target comparison in the controller.

query-graphs/src/ui/animated-layout.ts
vogelsgesang: 🤖 Addressed in `2a04393`. Exiting nodes now resolve their destination through the same anchor mechanism, including after retargeting; the stored `exitPosition` was removed.

query-graphs/src/ui/animated-layout.ts
vogelsgesang: ok, do as discussed; keep `sameLayoutTarget` separate

vogelsgesang: 🤖 Addressed in `7d8a972`. `createLayoutInterpolator` now builds its node/edge maps and ID unions once per transition and reuses them for every frame; `sameLayoutTarget` remains separate in the controller as discussed.

query-graphs/src/ui/animated-layout.ts
```diff
@@ -0,0 +1,142 @@
...
+    x: number;
+    y: number;
+}
+
+export interface LayoutAnchor {
+    nodeId: string;
+    offset: Position;
+}
+
+export type TransitionAnchors = ReadonlyMap<string, LayoutAnchor>;
```
vogelsgesang, line 19: should `TransitionAnchors` be directly embeded into `AnimatedNode`?

query-graphs/src/ui/animated-layout.ts
```diff
@@ -0,0 +1,142 @@
...
+    const nodeIds = new Set([...fromNodes.keys(), ...toNodes.keys()]);
+    const fromEdges = new Map(from.edges.map((entry) => [entry.edge.id, entry]));
+    const toEdges = new Map(to.edges.map((edge) => [edge.id, edge]));
+    const edgeIds = new Set([...fromEdges.keys(), ...toEdges.keys()]);
+
+    return (progress) => {
+        const nodes: AnimatedNode[] = [];
+        for (const id of nodeIds) {
+            const start = fromNodes.get(id);
+            const target = toNodes.get(id);
```
vogelsgesang: can we index the nodes / edges on interpolator creation differently, so we don't need repeated lookups?

query-graphs/src/ui/animated-layout.ts
vogelsgesang: 🤖 I would keep anchors out of `AnimatedNode`: they describe one transition, while `AnimatedNode` persists across transitions, so embedding them would require carrying and clearing stale metadata. Instead, the interpolator factory can resolve each anchor once into a transition-local record containing the start and target positions.

query-graphs/src/ui/animated-layout.ts
vogelsgesang: 🤖 Yes. I recommend building arrays of node and edge transition records in `createLayoutInterpolator`, with positions, opacities, and entering/exiting state resolved up front. Each frame would then iterate those arrays directly, with no map lookups or anchor resolution. This also addresses the anchor-storage question without adding transition state to `AnimatedNode`.

vogelsgesang: 🤖 Addressed in `0f98d1e`. `createLayoutInterpolator` now resolves node and edge transitions, including handle-anchor positions, once when the interpolator is created. Each animation frame iterates those prepared records without repeated map lookups or anchor resolution.

query-graphs/src/ui/animated-layout.ts
vogelsgesang: 🤖 Implemented the transition-local approach in `0f98d1e`: anchors remain separate from persistent `AnimatedNode` state and are resolved once into the interpolator's prepared node-transition records.

query-graphs/src/ui/useAnimatedGraphLayout.ts
```diff
@@ -0,0 +1,464 @@
+import type {Dimensions, InternalNode, NodeChange} from "@xyflow/react";
```
vogelsgesang: add a top-level comment describing the intended usage

query-graphs/src/ui/useAnimatedGraphLayout.ts
```diff
@@ -0,0 +1,464 @@
...
+    nodeElement: HTMLElement;
+}
+
+/**
+ * Animates a synchronously applied change. List persistent nodes whose bodies
+ * may resize; entering and exiting nodes are inferred afterwards.
+ */
+export type AnimateGraphChange = (applyChange: () => void, resizingNodes?: readonly NodeResizeRequest[]) => void;
+
+export const subtreeHandleId = "subtree";
```
vogelsgesang: * document
* move hooks and contexts

query-graphs/src/ui/useAnimatedGraphLayout.ts
```diff
@@ -0,0 +1,455 @@
...
+/** Tests dimensions by value so unchanged measurements retain their map identity. */
+function sameDimensions(left: Dimensions | undefined, right: Dimensions): boolean {
+    return left?.width === right.width && left.height === right.height;
+}
+
+/**
+ * Reconciles dimensions reported by React Flow. Measured dimensions describe
+ * the current rendered frame; target dimensions drive the stable tree layout.
+ * Active body resizes therefore retain their target until the new final size
+ * is measured. Unchanged inputs preserve their map and state identities.
```
vogelsgesang: "Unchanged inputs preserve their map and state identities." - its unclear what "map and state identities" are. Rephrase.

query-graphs/src/ui/useAnimatedGraphLayout.ts
```diff
@@ -0,0 +1,455 @@
...
+
+/**
+ * Animates a graph change applied synchronously by `applyChange`. List
+ * persistent nodes whose bodies may resize; entering and exiting nodes are
+ * inferred from the resulting layout.
+ */
+export type AnimateGraphChange = (applyChange: () => void, resizingNodes?: readonly NodeResizeRequest[]) => void;
+
+/** Identifies the source handle from which changed subtrees emerge or recede. */
+export const subtreeHandleId = "subtree";
```
vogelsgesang: move after `useAnimateGraphChange`

query-graphs/src/ui/useAnimatedGraphLayout.ts
```diff
@@ -0,0 +1,455 @@
...
+    const handle = node?.internals.handleBounds?.source?.find((candidate) => candidate.id === handleId);
+    if (node === undefined || handle === undefined) return undefined;
+
+    const x = node.internals.positionAbsolute.x + handle.x + handle.width / 2;
+    const y = node.internals.positionAbsolute.y + handle.y + handle.height / 2;
+    return {nodeId: node.id, offset: {x: x - node.position.x, y: y - node.position.y}};
+}
+
+/** Indexes every node's parent once, including currently collapsed children. */
+function treeParents(tree: TreeDescription, nodeIds: Map<TreeNode, string>): ReadonlyMap<string, string> {
```
vogelsgesang: should this also be moved to animated-layout.ts? If so, do it

query-graphs/src/ui/useAnimatedGraphLayout.ts
vogelsgesang: 🤖 Addressed

query-graphs/src/ui/useAnimatedGraphLayout.ts
vogelsgesang: 🤖 Addressed

query-graphs/src/ui/useAnimatedGraphLayout.ts
vogelsgesang: 🤖 Addressed

query-graphs/src/ui/useAnimatedGraphLayout.ts
vogelsgesang: 🤖 Moved `treeParents` to `animated-layout.ts`; it is pure ancestry indexing used directly by `transitionAnchors`, while the hook now retains React and DOM coordination.

query-graphs/test/animated-layout.test.ts
```diff
@@ -0,0 +1,278 @@
...
+        ["left-child", {nodeId: "left-parent", offset: {x: 0, y: 30}}],
+        ["right-child", {nodeId: "right-parent", offset: {x: 0, y: 30}}],
+    ]);
+
+    const halfway = createLayoutInterpolator(start, target, anchors)(0.5);
+    assert.deepEqual(halfway.nodes.find((entry) => entry.node.id === "left-child")?.position, {x: 5, y: 65});
+    assert.deepEqual(halfway.nodes.find((entry) => entry.node.id === "right-child")?.position, {x: 105, y: 65});
+});
+
+test("transition anchors resolve deep changed subtrees in one ancestry pass", () => {
```
vogelsgesang: remove; implementation detail; no need to test

query-graphs/test/animated-layout.test.ts
```diff
@@ -0,0 +1,278 @@
...
+    createLayoutInterpolator,
+    refreshLayoutData,
+    sameLayoutTarget,
+    staticLayout,
+    transitionAnchors,
+} from "../src/ui/animated-layout";
+import type {GraphLayout, TransitionAnchors} from "../src/ui/animated-layout";
+import {graphAnimationDuration, graphAnimationProgress} from "../src/ui/animation-timing";
+import type {QueryGraphNode} from "../src/ui/QueryNode";
+import {measurePendingBodyResizes, reconcileDimensions} from "../src/ui/useAnimatedGraphLayout";
```
vogelsgesang: move test cases for those into a separate test file

query-graphs/src/ui/useAnimatedGraphLayout.ts
```diff
@@ -0,0 +1,458 @@
+/**
+ * Coordinates animated query-graph changes with React Flow. `QueryGraph` passes
+ * the returned nodes, edges, and `onNodesChange` callback to React Flow and
+ * exposes `animateGraphChange` to nodes. Callers wrap synchronous graph updates
+ * in that callback and identify persistent nodes whose bodies may resize.
+ *
+ * Pure layout transitions live in `animated-layout.ts`; this module coordinates
+ * React state, React Flow measurements, DOM body measurements, and animation
+ * frames.
```
vogelsgesang:
```suggestion
 * Coordinates animated query-graph changes with React Flow.
 * Pass the returned nodes, edges, and `onNodesChange` callback to React Flow and
 * set up an `AnimateGraphChangeContext.Provider`. Via `useAnimateGraphChange`,
 * you can then request animations.
 *
 * Pure layout transitions live in `animated-layout.ts`; this module coordinates
 * React state, React Flow measurements, DOM body measurements, and animation
 * frames.
```

query-graphs/test/animated-layout.test.ts
vogelsgesang: 🤖 Addressed

query-graphs/test/animated-layout.test.ts
vogelsgesang: 🤖 Addressed

query-graphs/src/ui/useAnimatedGraphLayout.ts
vogelsgesang: 🤖 Addressed

query-graphs/src/ui/useAnimatedGraphLayout.ts
```diff
@@ -0,0 +1,430 @@
...
+    nodeElement: HTMLElement;
+    bodyElement: HTMLElement;
+    bodyStart: Dimensions;
+    bodyTarget?: Dimensions;
+}
+
+/**
+ * Captures a body's current size before a graph change. `nodeElement` is a
+ * descendant of the containing React Flow node, whose body supplies the
+ * independently animated dimensions.
```
vogelsgesang:
```suggestion
 * descendant of the containing React Flow node, whose body supplies the
 * animated dimensions.
```

query-graphs/src/ui/useAnimatedGraphLayout.ts
```diff
@@ -0,0 +1,430 @@
...
+    expandedSubtrees: Record<string, boolean>,
+): GraphLayout & {
+    onNodesChange: (changes: NodeChange<QueryGraphNode>[]) => void;
+    animateGraphChange: AnimateGraphChange;
+} {
+    // React Flow services used to resolve handles and fit the initial graph.
+    const {fitView, getInternalNode} = useReactFlow<QueryGraphNode>();
+
+    // QueryGraph remounts this hook for each tree, so measurements belong only
+    // to the graph instance that collected them.
```
vogelsgesang: remove comment; not relevant without the history of this PR

query-graphs/src/ui/QueryNode.css
```diff
@@ -1,24 +1,41 @@
...
+    &.qg-expanded,
+    &.qg-resizing {
         cursor: pointer;
         background: white;
         border-color: hsl(0,0%, 80%);
         box-shadow: 1px 2px 5px hsl(0,0%,90%);
     }
 }
 
+.qg-graph-node.qg-resizing {
```
vogelsgesang: nest inside existing `.qg-graph-node` section

PR #173, opened 2026-09-09
W-23900223 : Hyper Query Plan - Replace raw property dumps with tidy node rows.

rbrahmachary:
Hyper Query Plan - Replace raw property dumps with tidy node rows.
  Node bodies previously dumped debug-name, statistics, and sqlpos verbatim via forceToString, producing noisy JSON blobs. Surface the useful figures as clean, curated rows instead.

  - Replace the raw dump with curated rows: table-or-alias (from debug-name.value), estimated-rows, output-rows, processed-rows, rows-matching, memory-bytes, cpu-cycles, execution-time
  - New : Add a grouped table-metadata block (identifier / partitioned-by /sort-order) for lakehouse scans, rendered as a header plus sub-items
  - Show memory-bytes in binary units via a new formatBytes helper (KiB/MiB/GiB), clearer than the decimal formatMetric
  - Display row order: identity/volume first, operator identifiers last
  - Echo the runtime hotspot flag onto the cpu-cycles row, reusing the node heat color so the tint intensity matches the label
  - Add getStatistic / getEstimatedRows / getActualRows helpers that read statistics and fall back to legacy analyze / cardinality
  - Add a firstNumber helper that skips non-numeric values instead of masking a real number further down the fallback chain
  - Reuse those helpers for the edge label, hotspot heatmap, and errored-icon detection, keeping tooltip figures and visual cues consistent (including on legacy-format plans)
  - Extract the hotspot and cardinality-mismatch magic numbers into constants
  - Add groupedProperties / highlightedProperties to TreeNode and the grouped-header / highlighted-row rendering in QueryNode

Query Graph node rows:-
Screenshot 2026-09-09 at 4 47 13 PM
Screenshot 2026-09-09 at 4 47 16 PM
Screenshot 2026-09-09 at 4 47 33 PM

### DIFF
--- a/query-graphs/src/loaders/hyper.ts
+++ b/query-graphs/src/loaders/hyper.ts
@@ -23,7 +23,7 @@ We transform a Hyper JSON tree into a query-graphs tree using the following heur
 import type {TreeNode, TreeDescription, Crosslink, IconName} from "../tree-description";
 import {allChildren} from "../tree-description";
 import type {Json, JsonObject} from "./loader-utils";
-import {forceToString, tryToString, formatMetric, hasOwnProperty, tryGetPropertyPath} from "./loader-utils";
+import {forceToString, tryToString, formatMetric, formatBytes, hasOwnProperty, tryGetPropertyPath} from "./loader-utils";
 
 // A categorical color palette for execution pipelines (the Tableau 20 colors).
 // The ten saturated base hues come first, then their lighter companions, so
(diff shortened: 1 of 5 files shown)


query-graphs/src/tree-description.ts
```diff
@@ -28,6 +28,11 @@ export type TreeNode = {
     iconColor?: string;
     // Rendered in the tooltip
     properties?: Map<string, string>;
+    // Property keys whose value is newline-separated `label: value` sub-lines (e.g. `table-metadata`); the UI
+    // renders them as a header row plus one sub-item per line.
+    groupedProperties?: Set<string>;
+    // Property keys whose row is highlighted (e.g. `cpu-cycles` on a hotspot), echoing the node-label flag.
+    highlightedProperties?: Set<string>;
```
vogelsgesang, line 35: why choose 3 different properties instead of a new PropertyEntry?

```suggestion
export type PropertyEntry = {
   value: string;
   highlighted: bool;
   grouped: bool;
};

    // Rendered in the tooltip
    properties?: Map<string, PropertyEntry>;
```

query-graphs/src/tree-description.ts
```diff
@@ -28,6 +28,11 @@ export type TreeNode = {
     iconColor?: string;
     // Rendered in the tooltip
     properties?: Map<string, string>;
+    // Property keys whose value is newline-separated `label: value` sub-lines (e.g. `table-metadata`); the UI
+    // renders them as a header row plus one sub-item per line.
+    groupedProperties?: Set<string>;
```
vogelsgesang, line 33: why encode "label: value" into a single string? Why not rely on type safety?

```
export type PropertyEntry = {
   value: string | Map<string, PropertyEntry>;
   highlighted: bool;
};
```

As soon as we have this representation, we can actually render arbitrary nested JSON structures as nested properties. We might even consider making them "expandable" / "collapsible" in the UI.

Can we now simply map the complete "statistics" JSON entry into a nested `PropertyEntry`? Can we replace all `forceToString` calls with a `jsonTreeAsNestedPropertyTree` converter

Maybe also add a `defaultVisible`, so we get the relevant ones by default, but you can still access everything if you want

query-graphs/src/loaders/hyper.ts
```diff
@@ -59,6 +59,66 @@ function pipelineColor(index: number): string {
...
+}
+
+// Measured output rows; old ANALYZE'd plans expose it as `analyze.tuple-count`.
+function getActualRows(rawNode: Json): number | undefined {
+    return firstNumber(getStatistic(rawNode, "output-rows"), tryGetPropertyPath(rawNode, ["analyze", "tuple-count"]));
+}
+
+// Reorder the property map to `front` keys first, then any keys not named in `front`/`tail` (in their existing
+// order), then `tail` keys last (in the given order). Missing keys are skipped.
+function reorderProperties(properties: Map<string, string>, front: string[], tail: string[] = []): void {
```
vogelsgesang, line 104: general utility; move out of hyper.ts into loader_utils

PR #176, opened 2026-09-18
Extract reusable plan conversion infrastructure

vogelsgesang:
I am planning to add support for additional plan formats (Umbra,
CedarDB, DuckDB, ...). Most JSON plan formats need the same basic
conversion from arbitrary JSON into a tree, but differ in how they
recognize operators, name and order nodes, collapse auxiliary data,
and decorate the result.

Hence, this PR extracts Hyper's JSON-to-tree conversion into a
configurable `decorated-json-tree` helper. The generic JSON, Hyper,
and Postgres loaders now share the same gracefully degrading
conversion: scalar fields become tooltip properties, nested values
remain visible, and format-specific configuration adds semantic
names, icons, structural children, metrics, and crosslinks.

While at it, this PR extracts format-independent post-processing for
crosslinks, relative edge widths, node coloring, and Hyper pipeline
coloring. It also gives Hyper and Postgres dedicated tests for their
format-specific behavior while retaining hash-based regression
coverage for every example plan.


query-graphs/src/loaders/postgres.ts
vogelsgesang: import from `tree-postprocessing` instead of redeclaring. Probably same in hyper.ts

query-graphs/src/loaders/tree-postprocessing.ts
```diff
@@ -0,0 +1,210 @@
+import type {Crosslink, IconName, TreeNode} from "../tree-description";
+import {allChildren, visitTreeNodes} from "../tree-description";
+import type {Json, JsonObject} from "./loader-utils";
+
+export interface NodeRenderingConfig {
```
vogelsgesang: wrong file? Belongs into `adaptive-plan-tree`?

query-graphs/src/loaders/tree-postprocessing.ts
```diff
@@ -0,0 +1,210 @@
...
+function pipelineColor(index: number): string {
+    return PIPELINE_PALETTE[index % PIPELINE_PALETTE.length];
+}
+
+export interface RawPipeline {
+    id: number;
+    operatorIds: number[];
+}
+
+export function parsePipelines(
```
vogelsgesang: `parsePipelines` seems to be loader-specific and should remain in the specific loaders?

query-graphs/src/loaders/postgres.ts
vogelsgesang:
```suggestion
// Convert Postgres JSON to a TreeDescription
```

query-graphs/src/loaders/tree-postprocessing.ts
```diff
@@ -0,0 +1,210 @@
...
+    crosslinks: UnresolvedCrosslink[];
+    edgeWidths: {node: TreeNode; width: number}[];
+    runtimes: {node: TreeNode; time: number}[];
+}
+
+export function newConversionState(): ConversionState {
+    return {crosslinks: [], edgeWidths: [], runtimes: []};
+}
+
+export function buildIdMap(root: TreeNode, keys: readonly string[]): Map<string, TreeNode> {
```
vogelsgesang: why is this used by Hyper but not Postgres? Should we unify them somehow?

query-graphs/src/loaders/postgres.ts
vogelsgesang: 🤖 Addressed in e1305e4: Postgres now imports the shared  type, and Hyper's adaptive state uses it as well.

query-graphs/src/loaders/tree-postprocessing.ts
vogelsgesang: 🤖 Addressed in e1305e4:  and the conversion state now live in .

query-graphs/src/loaders/tree-postprocessing.ts
vogelsgesang: 🤖 Addressed in e1305e4: pipeline parsing is back in ; only the format-independent visualization lives in .

query-graphs/src/loaders/postgres.ts
vogelsgesang: 🤖 Addressed in a494581.

query-graphs/src/loaders/tree-postprocessing.ts
vogelsgesang: 🤖 Addressed in b7ea2fc: Postgres and Hyper now both build their ID map from the converted tree before resolving crosslinks.

query-graphs/src/loaders/hyper.ts
```diff
@@ -174,436 +109,142 @@ const legacyNodeTags: Record<string, string> = {
...
+            continue;
+        }
         const id = entry["id"];
         const operators = entry["operators"];
-        if (typeof id !== "number" || !Array.isArray(operators)) continue;
-        const operatorIds = operators.filter((o): o is number => typeof o === "number");
+        if (typeof id !== "number" || !Array.isArray(operators)) {
+            continue;
+        }
+        const operatorIds = operators.filter((operatorId): operatorId is number => typeof operatorId === "number");
```
vogelsgesang: unnecessary churn; remove

docs/PlanFormatsAndLoaders.md
```diff
@@ -16,7 +16,9 @@ Everything downstream (layout, rendering, interaction) is shared across formats.
 | Generic XML | `xml.ts` | XML | fallback — renders any XML as a tree |
 
 The Postgres and Hyper loaders understand plan semantics: they choose icons, order and collapse children, label edges with cardinalities, and color nodes by runtime.
-The generic JSON and XML loaders map the input structure literally and act as catch-all fallbacks.
+The generic JSON loader renders scalar fields as tooltip properties and nested values as tree nodes, without format-specific decorations or collapsed children.
+The generic XML loader maps the input structure literally.
+Both act as catch-all fallbacks.
```
vogelsgesang: don't change this; the exact way the JSON is mapped is an implementation detail

query-graphs/src/loaders/hyper.ts
```diff
@@ -174,436 +109,156 @@ const legacyNodeTags: Record<string, string> = {
...
-    // ever sees colors).
-    const nodePipelines = new Map<TreeNode, ResolvedPipeline[]>();
-    for (const p of resolved) {
-        for (const node of p.nodes) {
-            const list = nodePipelines.get(node) ?? [];
-            list.push(p);
-            nodePipelines.set(node, list);
-        }
+    const root = convertDecoratedJsonNode(node, "result", state, hyperConfig);
+    if (Array.isArray(root)) {
```
vogelsgesang: should be handled inside `convertDecoratedJsonNode` and should be handled more gracefully

query-graphs/src/loaders/hyper.ts
```diff
@@ -174,436 +109,156 @@ const legacyNodeTags: Record<string, string> = {
...
-        crosslinks: [],
-        edgeWidths: [],
-        runtimes: [],
-        metadata: new Map<string, string>(),
-    } as ConversionState;
-    // Check if the query failed
-    const errorMsg = tryGetPropertyPath(node, ["statistics", "error", "message", "original"]);
-    if (errorMsg) {
-        conversionState.metadata.set("Error", forceToString(errorMsg));
+function convertOptimizerSteps(node: Json): TreeDescription | undefined {
```
vogelsgesang, line 212: why did this function become more complicated now?

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang, line 1: this file is underdocumented. Add comments, especially regarding the intended usage and design principles

query-graphs/test/loaders.test.ts
```diff
@@ -72,6 +72,9 @@ test("dispatcher falls back to the generic JSON loader", () => {
     for (const json of ['{"unrecognized":{"value":42}}', '[{"name":"Alice","children":[]}]', '{"operator":{}}']) {
         assert.equal(loadPlanFromText(json).format, "json", json);
     }
+
+    const tree = loadPlanFromText('{"nested":{"value":42},"items":[{"value":1}]}').tree;
```
vogelsgesang, line 35: what is the intended test coverage here? Document in code comment

query-graphs/src/loaders/tree-postprocessing.ts
```diff
@@ -0,0 +1,56 @@
...
+    for (const link of crosslinks) {
+        const target = operatorsById.get(link.targetId);
+        if (target !== undefined) {
+            resolved.push({source: link.source, target});
+        }
+    }
+    return resolved;
+}
+
+export function colorRelativeExecutionTime(
```
vogelsgesang: generalize into `colorRelativeNumber`; rename `runtimes` accordingly

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in 1d50a49.

query-graphs/src/loaders/tree-postprocessing.ts
```diff
@@ -0,0 +1,56 @@
...
+    totalTime = runtimes.reduce((sum, runtime) => sum + runtime.time, 0),
+): void {
+    for (const runtime of runtimes) {
+        const relativeExecutionRatio = runtime.time / totalTime;
+        const lightness = (95 + (72 - 95) * relativeExecutionRatio).toFixed(3);
+        runtime.node.nodeColor = relativeExecutionRatio >= 0.05 ? `hsl(309, 84%, ${lightness}%)` : undefined;
+    }
+}
+
+export function setEdgeWidths(edgeWidths: {node: TreeNode; width: number}[]): void {
```
vogelsgesang: rename to `setRelativeEdgeWidths`

docs/PlanFormatsAndLoaders.md
vogelsgesang: 🤖 Addressed in 51a8129.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in 8c30a44. The converter now always returns one root node and wraps array roots instead of making callers reject them.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in abb1d66 by restoring the original straightforward validation flow.

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang: 🤖 Addressed in 4cdc7a8 with module-level design principles and documentation for the configuration, state, and root-wrapping contracts.

query-graphs/src/loaders/json.ts
```diff
@@ -3,55 +3,32 @@
...
 export const jsonPlanLoader: PlanLoader<Json> = {
     format: "json",
     matches: () => true,
     load(json) {
-        const root = {name: "root", children: convertChildren(json)};
+        const state = createDecoratedJsonTreeState();
+        const root = convertDecoratedJsonNode(json, "root", state, jsonTreeConfig);
+        if (!root.name) {
+            root.name = "root";
+        }
```
vogelsgesang: should this live inside `convertDecoratedJsonNode`?

query-graphs/test/loaders.test.ts
vogelsgesang: 🤖 Addressed in 02696b6.

query-graphs/src/loaders/tree-postprocessing.ts
vogelsgesang: 🤖 Addressed in 99ded4e.

query-graphs/src/loaders/pipeline-coloring.ts
```diff
@@ -0,0 +1,116 @@
...
+function pipelineColor(index: number): string {
+    return PIPELINE_PALETTE[index % PIPELINE_PALETTE.length];
+}
+
+export interface RawPipeline {
+    id: number;
+    operatorIds: number[];
+}
+
+export function assignPipelineColors(
```
vogelsgesang, line 45: the original code contained comments; preserve them

query-graphs/src/loaders/tree-postprocessing.ts
vogelsgesang: 🤖 Addressed in 93cdd92.

query-graphs/src/loaders/hyper.ts
```diff
@@ -587,23 +211,42 @@ function convertOptimizerSteps(node: Json): TreeDescription | undefined {
...
+}
+
+function isHyperPlanRoot(json: Json): json is JsonObject {
+    return (
+        typeof json === "object" &&
+        !Array.isArray(json) &&
+        json !== null &&
+        (typeof json["operator"] === "string" || typeof json["expression"] === "string")
+    );
 }
```
vogelsgesang: why move those functions around? seems like unnecessary churn?

query-graphs/src/loaders/json.ts
vogelsgesang: 🤖 Addressed in 1cd5809. The shared converter now guarantees the requested root name.

query-graphs/src/loaders/pipeline-coloring.ts
vogelsgesang: 🤖 Addressed in 39dd768 by restoring the original pipeline-coloring comments.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in ed77f6d by restoring the original helper ordering.

query-graphs/src/loaders/decorated-json-tree.ts
```diff
@@ -0,0 +1,234 @@
+/**
+ * Configurable JSON-to-tree conversion shared by the generic JSON loader and
+ * loaders that understand particular JSON plan formats.
+ *
+ * The baseline conversion is deliberately loss-tolerant: scalar object members
+ * become tooltip properties, while nested objects and arrays remain visible as
+ * tree nodes. Format loaders decorate that baseline through configuration; they
+ * can recognize tagged nodes, choose names and icons, order or flatten children,
+ * collapse auxiliary data, and extract metrics and crosslinks. Unknown fields
+ * continue through the baseline conversion instead of being discarded.
```
vogelsgesang:
```suggestion
 * The baseline conversion is gracefully degrading: scalar object members
 * become tooltip properties, while nested objects and arrays remain visible as
 * tree nodes. Format loaders decorate that baseline through configuration; they
 * can recognize tagged nodes, choose names and icons, order or flatten children,
 * collapse auxiliary data, and extract metrics and crosslinks. Unknown fields
 * continue through the baseline conversion instead of being discarded.
```

query-graphs/src/loaders/decorated-json-tree.ts
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * Conversion only records data needed by later whole-tree passes. Resolving
+ * crosslinks, scaling edges, and coloring relative values stays outside this module.
```
vogelsgesang: remove this part


```suggestion
```

query-graphs/src/loaders/decorated-json-tree.ts
```diff
@@ -0,0 +1,234 @@
...
+
+export interface DecoratedJsonTreeConfig {
+    /** Choose presentation details for an object recognized by one of `nodeTypeKeys`. */
+    getRenderingConfig(nodeTypeKey: string, tag: string, rawNode: JsonObject): NodeRenderingConfig;
+    /** Object keys whose scalar values identify a semantic node type, in precedence order. */
+    nodeTypeKeys: readonly string[];
+    /** Keep these values in the tooltip even when they are objects or arrays. */
+    alwaysPropertyKeys: readonly string[];
+    /** Show these children first and omit their otherwise redundant key wrapper. */
+    fixedChildOrder: readonly string[];
```
vogelsgesang: order by importance:
1. nodeTypeKeys
2. fixedChildOrder
3. alwaysPropertyKeys
4. getRenderingConfig

Here, and in all users of this

vogelsgesang: 🤖 Addressed in 4570792.

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang: 🤖 Addressed in 4570792.

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang: 🤖 Addressed in 4570792.

query-graphs/test/loaders.test.ts
```diff
@@ -68,10 +68,40 @@ test("dispatcher recognizes the Postgres and Hyper examples", () => {
     }
 });
 
+test("Postgres decorates the shared JSON tree with plan-specific semantics", () => {
```
vogelsgesang: structure this test file more clearly.

First, there should be a section re general "loader behavior" (fallback, etc.), then the pinned hashes of all plans. The Postgres / Hyper / ... specific tests should be in a `postgres.test.ts` / `hyper.test.ts`. While at it, also add targetted test cases for Hyper and Postgres' features

query-graphs/src/loaders/postgres.ts
```diff
@@ -82,149 +67,32 @@ function getOperatorRendering(operatorType: string, properties: Map<string, stri
...
-                listOfObjects.push(innerNode);
-            }
-        }
-        return listOfObjects;
-    }
-    throw new InvalidPlanError("postgres");
-}
+const postgresConfig: DecoratedJsonTreeConfig = {
+    nodeTypeKeys: ["Node Type"],
+    fixedChildOrder: ["Plan", "Plans"],
```
vogelsgesang: really needed? Do they ever occur at the same time?

Also, don't we need to define an order for "left", "right", etc? Or is Postgres not using that?

query-graphs/src/loaders/decorated-json-tree.ts
```diff
@@ -0,0 +1,233 @@
...
+            const node = Array.isArray(converted) ? {children: converted} : converted;
+            if (!node.name) {
+                node.name = name;
+            }
+            return node;
+        });
+    }
+
+    if (typeof rawNode !== "object" || rawNode === null) {
+        throw new Error("Invalid query plan");
```
vogelsgesang: should throw a different error type; see `types.ts`

query-graphs/src/loaders/decorated-json-tree.ts
```diff
@@ -0,0 +1,233 @@
...
+    const properties = new Map<string, string>();
+    const {nodeTypeKey, nodeTag} = classifyNode(rawNode, config);
+
+    for (const key of config.alwaysPropertyKeys) {
+        if (hasOwnProperty(rawNode, key)) {
+            properties.set(key, forceToString(rawNode[key]));
+        }
+    }
+
+    for (const key of orderedKeys(rawNode, nodeTypeKey, config)) {
```
vogelsgesang, line 165: if I rememeber correctly, the Hyper version of this algorithm contained some explanatory comments. Keep them

query-graphs/src/loaders/postgres.ts
vogelsgesang: 🤖 Both entries are needed here as flattening markers, rather than because they can occur together: `Plan` contains the top-level operator, while `Plans` contains an operator's ordered inputs. Their relative order is therefore irrelevant. PostgreSQL JSON plans encode left/right input order by array position in `Plans`; they do not use `left`/`right` keys.

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang: 🤖 Addressed in 1db04a4.

query-graphs/src/loaders/postgres.ts
vogelsgesang: makes sense. Can we find a more speaking name for `fixedChildOrder` then? `flattenedChildren` maybe? Ideate.

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang: 🤖 Addressed in 19f6fc6 by restoring the conversion rationale around ordering, adaptive rendering, flattening, array collapsing, naming, and post-conversion decorations.

query-graphs/test/loaders.test.ts
vogelsgesang: 🤖 Addressed in 14d179e. General dispatch behavior now comes first in `loaders.test.ts`, digest pinning comes last, and Hyper/Postgres recognition and targeted feature coverage live in dedicated test files.

query-graphs/src/loaders/postgres.ts
vogelsgesang: 🤖 Agreed—the name should express more than ordering. Options:

- `structuralChildKeys`: my preference; these are the format's actual tree edges, shown first in the listed order and without a key wrapper.
- `flattenedChildKeys`: clearest about the wrapper removal, but does not advertise that array order controls display order.
- `flattenedChildOrder`: captures both behaviors, though it reads more like a sorting function than a list of keys.
- `primaryChildKeys`: conveys prominence, but not flattening.

I recommend `structuralChildKeys`, with the field comment explicitly stating that list order is preserved and wrappers are omitted. `flattenedChildKeys` would also be reasonable if you prefer the mechanical behavior in the name.

query-graphs/src/loaders/decorated-json-tree.ts
```diff
@@ -0,0 +1,253 @@
...
+        appendChild(target, converted, key, config.fixedChildOrder.includes(key), collapse);
+    }
+
+    // Determine format-specific rendering and the most meaningful available name.
+    const renderingConfig =
+        nodeTypeKey !== undefined && nodeTag !== undefined ? config.getRenderingConfig(nodeTypeKey, nodeTag, rawNode) : {};
+    const displayName =
+        config.getDisplayName?.(rawNode) ??
+        (renderingConfig.displayNameKey === undefined ? undefined : properties.get(renderingConfig.displayNameKey)) ??
+        properties.get("name") ??
```
vogelsgesang: we shouldn't need the `properties.get("name")` fallback, do we?

package.json
```diff
@@ -8,8 +8,8 @@
     "scripts": {
         "lint": "eslint . --fix",
         "lint-test": "eslint .",
-        "test": "node --import=tsx --test --test-isolation=none query-graphs/test/loaders.test.ts",
-        "test:update": "node --import=tsx --test --test-isolation=none --test-update-snapshots query-graphs/test/loaders.test.ts"
+        "test": "node --import=tsx --test --test-isolation=none query-graphs/test/loaders.test.ts query-graphs/test/postgres.test.ts query-graphs/test/hyper.test.ts",
+        "test:update": "node --import=tsx --test --test-isolation=none --test-update-snapshots query-graphs/test/loaders.test.ts query-graphs/test/postgres.test.ts query-graphs/test/hyper.test.ts"
```
vogelsgesang: can we auto-enumerate all `*.test.ts` files somehow easily?

query-graphs/test/loaders.test.ts
```diff
@@ -98,12 +60,8 @@ test("dispatcher recognizes Tableau XML and falls back to generic XML", () => {
 });
 
 test("dispatcher can force a registered loader", () => {
-    const hyper = readFileSync(path.join(examplesRoot, "hyper/tablescan-analyze.plan.json"), "utf8");
-    assert.equal(loadPlanFromText(hyper, {format: "hyper"}).format, "hyper");
-    assert.equal(loadPlanFromText(hyper, {format: "json"}).format, "json");
-    assert.equal(loadPlanFromText(hyper, {format: "json"}).tree.root.name, "root");
-
-    assert.equal(loadPlanFromText('{"operator":{}}', {format: "hyper"}).format, "hyper");
```
vogelsgesang, line 106: why did we remove that coverage? We should still test that we can force specific loaders such as hyper, postgres, etc.?

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang: 🤖 The behavior is needed for generic JSON: for example, `[{"name":"Alice","children":[]}]` should render as `Alice`, not `root.0`. But I agree it does not have to be an implicit shared-converter fallback. A cleaner alternative is to make the JSON loader's `getDisplayName` explicitly read `name` (and preserve Hyper's existing `name` fallback in its own callback), then remove `properties.get("name")` here. I recommend that separation; let me know if you agree and I’ll make it a follow-up.

package.json
vogelsgesang: 🤖 Addressed in df72a48. Node 24 expands the quoted `query-graphs/test/*.test.ts` pattern directly, so new test files are discovered without editing the scripts.

query-graphs/src/loaders/postgres.ts
vogelsgesang: use `structuralChildKeys`

query-graphs/test/loaders.test.ts
vogelsgesang: 🤖 Addressed in 4cc2d94. The dispatcher test again forces both Hyper and Postgres explicitly, and verifies that the same inputs can instead be forced through the generic JSON loader.

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang: do it

query-graphs/src/loaders/postgres.ts
vogelsgesang: 🤖 Addressed in 36f5eaf: renamed the configuration to `structuralChildKeys` and clarified that its listed order is preserved while redundant key wrappers are omitted.

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang: 🤖 Addressed in 9b6770c. The shared converter no longer assumes a `name` property; the generic JSON and Hyper loaders now opt into that fallback explicitly.

PR #180, opened 2026-09-18
Initial support for DuckDB plans

vogelsgesang:
This commit adds support for DuckDB plans. The renderer shows meaningful operator names and icons, estimated and actual cardinalities, execution timing, profile metadata, and CTE and delimiter-scan relationships. DuckDB CAPS_LOCK operator labels are displayed in lowercase as this is easier on the eyes.

The loader handles JSON EXPLAIN plans, analyzed profiles, and logical/optimized/physical stage dumps. It builds on the shared decorated JSON converter, uses DuckDB mandatory profile fields for narrow recognition, and degrades malformed or unknown content gracefully.

We also extended the shared `decorated-json-tree.ts` to support flattening of nested property objects and actual-only cardinalities. Actual-only values render as `actual/?`, keeping them distinct from estimate-only labels.

DuckDB uses the same narrow node signature to recognize plain, analyzed, and optimizer-stage plans, while leaving envelope loading permissive when the format is selected explicitly.

As a drive-by cleanup, this commit also removes the unused `jsonToStringMap` helper.


query-graphs/src/loaders/decorated-json-tree.ts
```diff
@@ -207,16 +225,19 @@ function convertDecoratedJsonValue(
...
-        const edgeWidth = actualCardinality ?? estimatedCardinality;
+    const actualCardinality = config.getActualCardinality?.(rawNode);
+    const edgeWidth = actualCardinality ?? estimatedCardinality;
+    if (edgeWidth !== undefined) {
         state.edgeWidths.push({node: convertedNode, width: edgeWidth});
         convertedNode.edgeLabel =
             actualCardinality === undefined
-                ? formatMetric(estimatedCardinality)
-                : `${formatMetric(actualCardinality)}/${formatMetric(estimatedCardinality)}`;
+                ? formatMetric(edgeWidth)
```
vogelsgesang: why `edgeWidth`? Directly using `estimatedCardinality` would be more readable?

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,274 @@
...
+import type {Json, JsonObject} from "./loader-utils";
+import {forceToString, hasOwnProperty, hasSubObject, isJsonObject, tryToString} from "./loader-utils";
+import {buildIdMap, colorRelativeNumber, resolveCrosslinks, setRelativeEdgeWidths} from "./tree-postprocessing";
+import {InvalidPlanError, type PlanLoader} from "./types";
+
+function getExtraInfo(rawNode: JsonObject): JsonObject | undefined {
+    return hasSubObject(rawNode, "extra_info") ? rawNode["extra_info"] : undefined;
+}
+
+function parseNumber(value: Json | undefined): number | undefined {
```
vogelsgesang: move to loader utils

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,274 @@
...
+        return Number.isFinite(value) ? value : undefined;
+    }
+    if (typeof value === "string" && value.trim() !== "") {
+        const parsed = Number(value);
+        return Number.isFinite(parsed) ? parsed : undefined;
+    }
+    return undefined;
+}
+
+function optionalString(value: Json | undefined): string | undefined {
```
vogelsgesang: do we really need this helper? Can we generalize it into loader utils? Where else can we reuse it?

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,274 @@
...
+    return (
+        optionalString(rawNode["operator_type"]) ??
+        optionalString(rawNode["operator_name"]) ??
+        optionalString(rawNode["name"]) ??
+        "unknown"
+    );
+}
+
+function getIcon(rawNode: JsonObject): IconName | undefined {
+    const operatorType = getOperatorType(rawNode).toUpperCase();
```
vogelsgesang: here and other places: Avoid unnecessary upper-casing

query-graphs/src/loaders/decorated-json-tree.ts
vogelsgesang: 🤖 Addressed. Renamed it to `cardinality`; the separate value is needed because actual cardinality takes precedence over the estimate for edge sizing.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. Moved the finite number/string parser to `loader-utils` as `tryToNumber`.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. Generalized this as `tryToNonNullString` in `loader-utils` and reused it in both DuckDB and Postgres.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. DuckDB operator and join types are now matched in their canonical emitted casing; the only remaining case conversion is the intentional CAPS_LOCK display-name heuristic.

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,263 @@
...
+import type {Json, JsonObject} from "./loader-utils";
+import {forceToString, hasOwnProperty, hasSubObject, isJsonObject, tryToNonNullString, tryToNumber} from "./loader-utils";
+import {buildIdMap, colorRelativeNumber, resolveCrosslinks, setRelativeEdgeWidths} from "./tree-postprocessing";
+import {InvalidPlanError, type PlanLoader} from "./types";
+
+function getExtraInfo(rawNode: JsonObject): JsonObject | undefined {
+    return hasSubObject(rawNode, "extra_info") ? rawNode["extra_info"] : undefined;
+}
+
+function normalizeOperatorName(name: string): string {
```
vogelsgesang: combine with `getOperatorType`

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,263 @@
...
+    return undefined;
+}
+
+function getDisplayName(rawNode: JsonObject): string {
+    const name = normalizeOperatorName(
+        tryToNonNullString(rawNode["operator_name"]) ??
+            tryToNonNullString(rawNode["name"]) ??
+            tryToNonNullString(rawNode["operator_type"]) ??
+            "unknown",
+    );
```
vogelsgesang: use `getOperatorType`

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,263 @@
...
+    structuralChildKeys: ["children"],
+    alwaysPropertyKeys: [],
+    flattenPropertyObjectKeys: ["extra_info"],
+    getRenderingConfig(_nodeTypeKey, _tag, rawNode) {
+        return {icon: getIcon(rawNode)};
+    },
+    getDisplayName,
+    getCrosslinkTarget(rawNode) {
+        const operatorType = getOperatorType(rawNode);
+        if (operatorType.includes("CTE_SCAN")) {
```
vogelsgesang: why "includes" instead of `=`?

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,263 @@
...
+        if (joinType?.includes("RIGHT")) return "right-join-symbol";
+        if (joinType?.includes("FULL")) return "full-join-symbol";
+        return "inner-join-symbol";
+    }
+    if (operatorType.includes("GROUP_BY") || operatorType.includes("AGGREGATE")) return "groupby-symbol";
+    if (operatorType.includes("ORDER_BY") || operatorType.includes("SORT") || operatorType === "TOP_N") return "sort-symbol";
+    if (operatorType.includes("FILTER") || operatorType === "LIMIT") return "filter-symbol";
+    if (
+        operatorType.includes("CTE") ||
+        operatorType.includes("DELIM_SCAN") ||
```
vogelsgesang: why "includes" instead of `=`? Justify each of the `includes`

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. Operator selection and CAPS_LOCK normalization now live together in `getOperatorType`.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. `getDisplayName` now uses `getOperatorType`.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. CTE scans now use a suffix match specifically to cover both `CTE_SCAN` and recursive `REC_CTE_SCAN`; the broad substring match is gone.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. Removed every `includes` check. Exact comparisons handle individual operators; suffix checks are limited to DuckDB families with implementation prefixes, such as `HASH_JOIN`, `PIECEWISE_MERGE_JOIN`, and `PERFECT_HASH_GROUP_BY`.

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,263 @@
...
+// DuckDB's JSON renderer wraps each plan's single root in an array.
+function isSingletonArray(json: Json): json is [Json] {
+    return Array.isArray(json) && json.length === 1;
+}
+
+function hasAnalyzedEnvelope(json: Json): json is JsonObject & {children: [Json]} {
+    return isJsonObject(json) && typeof json["query_name"] === "string" && isSingletonArray(json["children"]);
+}
+
+function isExplainAnalyzeNode(value: Json): value is JsonObject & {children: Json[]} {
```
vogelsgesang: inline

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,263 @@
...
+function loadDuckDbPlan(json: Json): TreeDescription {
+    if (isSingletonArray(json) && isDuckNode(json[0])) {
+        return convertDuckPlan(json[0]);
+    }
+    if (hasAnalyzedEnvelope(json)) {
+        let root = json["children"][0];
+        if (isExplainAnalyzeNode(root)) {
+            if (!isSingletonArray(root["children"])) {
+                throw new InvalidPlanError("duckdb");
+            }
```
vogelsgesang: don't throw; be permissive

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,263 @@
...
+        const value = json[key];
+        if (key === "extra_info" && isJsonObject(value) && Object.keys(value).length === 0) {
+            continue;
+        }
+        metadata.set(key, forceToString(value));
+    }
+    return metadata;
+}
+
+function combinePlanStages(stages: [string, Json][]): TreeDescription {
```
vogelsgesang, line 204: should we pull this out and reuse across different loaders? Where would this best live? in tree-utils, as a generic `combineTrees` maybe?

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,263 @@
...
+            }
+            root = root["children"][0];
+        }
+        return convertDuckPlan(root, analyzeMetadata(json));
+    }
+    const stages = getPlanStages(json);
+    if (stages !== undefined) {
+        return combinePlanStages(stages);
+    }
+    throw new InvalidPlanError("duckdb");
```
vogelsgesang: restructure to be permissive. E.g., maybe use `return convertDuckPlan(json[0]);` as the fallback here and remove the initial `isSingletonArray(json) && isDuckNode(json[0])` branch?

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,263 @@
...
+    ["logical_opt", "optimized logical plan"],
+    ["physical_plan", "physical plan"],
+]);
+
+function getPlanStages(json: Json): [string, Json][] | undefined {
+    if (!isJsonObject(json) || !Array.from(stageNames.keys()).every((key) => hasOwnProperty(json, key))) {
+        return undefined;
+    }
+    const stages: [string, Json][] = [];
+    for (const [key, name] of stageNames) {
```
vogelsgesang: currenlty we drop unrecognized stages. We shouldn't do so. We should use the names only as hints and forward unknown stages as they are. Also, we don't actually need the pretty printing of the names here.

Probably, to detect if this is a multi-stage plan, we could check if at least one of the known keys exists?

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed by inlining the `EXPLAIN_ANALYZE` check in the loading path.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. A malformed `EXPLAIN_ANALYZE` wrapper is now rendered as-is instead of throwing.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. Explicit DuckDB loading now falls back to converting the first array entry (or the input itself), so the dedicated initial branch and terminal error are gone.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 Addressed. Any one known stage key now identifies the multi-stage envelope; all entries are retained in source order and rendered under their original keys, including unknown future stages.

query-graphs/src/loaders/duckdb.ts
vogelsgesang: 🤖 I would keep this local for now. The shared mechanics are only a short loop in two loaders, while the wrapper name and collapsed-child presentation are loader policy. A generic helper would either encode optimizer-stage semantics or need several presentation parameters. If a third use appears, `tree-description.ts` would be the clean home and `combineTreeDescriptions` a more precise name than `combineTrees`, since it composes roots plus crosslinks.

query-graphs/src/loaders/duckdb.ts
```diff
@@ -0,0 +1,241 @@
...
+        if (isDuckNode(root) && root["operator_type"] === "EXPLAIN_ANALYZE" && isSingletonArray(root["children"])) {
+            root = root["children"][0];
+        }
+        return convertDuckPlan(root, analyzeMetadata(json));
+    }
+    const stages = getPlanStages(json);
+    if (stages !== undefined) {
+        return combinePlanStages(stages);
+    }
+    return convertDuckPlan(Array.isArray(json) && json.length > 0 ? json[0] : json);
```
vogelsgesang, line 232:
```suggestion
    // Plans with multiple stages (`SET explain_output='all'`)
    const stages = getPlanStages(json);
    if (stages !== undefined) {
        return combinePlanStages(stages);
    }

    // ANALYZEd plans
    if (hasAnalyzedEnvelope(json)) {
        let root = json["children"][0];
        if (isDuckNode(root) && root["operator_type"] === "EXPLAIN_ANALYZE" && isSingletonArray(root["children"])) {
            root = root["children"][0];
        }
        return convertDuckPlan(root, analyzeMetadata(json));
    }

    // Normal plans
    return convertDuckPlan(Array.isArray(json) && json.length > 0 ? json[0] : json);
```

vogelsgesang: 🤖 Addressed exactly as suggested.

PR #182, opened 2026-09-18
Adapt loaders to current Hyper and Umbra plans

vogelsgesang:
Adapt the loaders to the updated plan formats from #183.

Hyper now renders the current result-sink and generic join and scan operators, reads runtime and error state from pipeline records, and attaches each complete statistics object to its pipeline driver as the pipeline-stats JSON property. Legacy operator aliases and direct-root plans remain supported through 2026-12-18, with dated removal TODOs on every compatibility path.

Umbra sourceLocation objects remain available as properties of their owning operators, expressions, and IU declarations instead of expanding into graph subtrees. Focused tests cover both format changes, while refreshed fixture hashes verify every committed plan.


plan-dumper/README.md
```diff
@@ -35,6 +35,8 @@ The available modes are:
   * MariaDB: its JSON `optimizer_trace`, which is a decision trace rather than a sequence of complete plans.
   * Postgres and Umbra currently skip this mode.
 * `pipelines` / `analyze-pipelines` → execution pipelines (Hyper only).
+  Current Hyper versions include pipelines automatically in fully optimized internal plans;
+  the separate modes retain stable example names and select whether to execute the plan.
```
vogelsgesang: remove; implementation detail; don't document in README

plan-dumper/dump-plans.py
```diff
@@ -342,11 +342,9 @@ def get_plan(sql, mode):
...
                     "external-analyze": "FORMAT JSON, ANALYZE, EXPAND_VIEWS true",
-                    "pipelines": (
-                        "FORMAT INTERNAL, PIPELINES, EXPAND_VIEWS true, EXPRESSIONS SQL"
-                    ),
+                    "pipelines": "FORMAT INTERNAL, EXPAND_VIEWS true, EXPRESSIONS SQL",
                     "analyze-pipelines": (
-                        "FORMAT INTERNAL, PIPELINES, ANALYZE, "
+                        "FORMAT INTERNAL, ANALYZE, "
                         "EXPAND_VIEWS true, EXPRESSIONS SQL"
                     ),
```
vogelsgesang: clean up the modes

```suggestion
                    "analyze-sql": "FORMAT INTERNAL, ANALYZE, EXPRESSIONS SQL",
```

query-graphs/src/loaders/hyper.ts
```diff
@@ -38,6 +41,26 @@ const nodeRenderingConfig: Record<string, NodeRenderingConfig> = {
...
+    // Inserts
+    "op:insert": {displayNameKey: "type"},
+    // Expressions
+    "exp:comparison": {displayNameKey: "mode"},
+    "exp:iu-ref": {displayNameKey: "iu"},
+    "exp:reference": {displayNameKey: "id"},
+};
+
+// TODO(2026-12-18): Remove rendering for operator tags replaced by generic result-sink, join, and scan operators.
+const legacyNodeRenderingConfig: Record<string, NodeRenderingConfig> = {
```
vogelsgesang: this mapping should only contain the legacy names which we will all remove at once. The names which are still in use should be in `nodeRenderingConfig`

query-graphs/src/loaders/hyper.ts
```diff
@@ -50,9 +73,6 @@ const nodeRenderingConfig: Record<string, NodeRenderingConfig> = {
     "op:left-mark-join": {crosslinkSourceKey: "magic"},
     "op:right-mark-join": {crosslinkSourceKey: "magic"},
     "op:early-probe": {icon: "filter-symbol", crosslinkSourceKey: "builder"},
-    // Various scans
```
vogelsgesang, line 53: keep the comments also here in this list

query-graphs/src/loaders/hyper.ts
vogelsgesang: do we actually need all entries below? Cnan't most of them be handled by the `legacyNodeTags` mapping below?

query-graphs/src/loaders/hyper.ts
```diff
@@ -145,9 +157,11 @@ const hyperConfig: DecoratedJsonTreeConfig = {
         return nodeTypeKey !== "operator";
     },
     isErrored(rawNode, metadata) {
+        // TODO(2026-12-18): Remove operator-level running state after the pipeline-statistics compatibility window.
         return metadata.has("Error") && tryGetPropertyPath(rawNode, ["statistics", "running"]) === true;
     },
```
vogelsgesang: let's remove this right away; and let's also remove the `isErrored` callback from the generic loader while at it

query-graphs/src/loaders/hyper.ts
```diff
@@ -277,13 +324,15 @@ function isHyperPlan(json: Json): boolean {
     if (hasPipelineEnvelope(json)) {
         return isHyperPlanRoot(json["tree"]);
     }
+    // TODO(2026-12-18): Stop auto-detecting fully optimized plans without the pipeline envelope.
```
vogelsgesang: what exactly do we mean with this comment? What exactly would we remove?

query-graphs/src/loaders/hyper.ts
```diff
@@ -277,13 +324,15 @@ function isHyperPlan(json: Json): boolean {
...
     return isOptimizerStepsPlan(json) || isHyperPlanRoot(json);
 }
 
 function loadHyperPlan(json: Json): TreeDescription {
     if (hasPipelineEnvelope(json)) {
         return convertHyperPlan(json["tree"], json["pipelines"]);
     }
+    // TODO(2026-12-18): Remove the direct-root fallback for fully optimized legacy plans.
```
vogelsgesang: what exactly do we mean with this comment? What exactly would we remove? why?

query-graphs/test/hyper.test.ts
```diff
@@ -70,6 +70,96 @@ test("Hyper leaves edges between disjoint pipeline memberships uncolored", () =>
     assert.equal(producer?.edgeColors, undefined);
 });
 
+test("Hyper applies pipeline-level runtime statistics", () => {
```
vogelsgesang:
```suggestion
test("Hyper applies pipeline-level runtime statistics to the pipeline driver", () => {
```

query-graphs/test/hyper.test.ts
```diff
@@ -70,6 +70,96 @@ test("Hyper leaves edges between disjoint pipeline memberships uncolored", () =>
...
+    assert.equal(scan?.properties?.get("pipeline-stats"), '{"cpu-cycles":25,"query-metrics":{"wall-clock":0.1},"running":false}');
+    assert.equal(failedPlan.metadata?.has("Pipeline statistics"), false);
+    assert.equal(failedPlan.root.iconColor, "red");
+    assert.equal(sort?.iconColor, "red");
+    assert.notEqual(scan?.iconColor, "red");
+    assert.equal(failedPlan.root.nodeColor, undefined);
+    assert.equal(scan?.nodeColor, undefined);
+});
+
+test("Hyper supports legacy plans through the compatibility window", () => {
```
vogelsgesang: in general, we don't test legacy plans; remove this test case

plan-dumper/README.md
vogelsgesang: 🤖 Addressed in #183 (`ee0a6ec`). Removed the obsolete README implementation detail and documented the single analyze-sql mode.

plan-dumper/dump-plans.py
vogelsgesang: 🤖 Addressed in #183 (`ee0a6ec`). Replaced both pipeline-specific modes with the suggested analyze-sql mode and regenerated the affected fixtures.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in `c89c61c`. The legacy rendering map now contains only the early-probe override; all current names remain in nodeRenderingConfig.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in `c89c61c`. Clarified that the dated cleanup will stop auto-detecting pre-2026-09-18 direct-root plans and require the {tree, pipelines} envelope.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in `c89c61c`. Clarified that the dated cleanup will stop loading pre-2026-09-18 direct-root plans and require the {tree, pipelines} envelope.

query-graphs/test/hyper.test.ts
vogelsgesang: 🤖 Addressed in `c89c61c`.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in `c89c61c`. Removed operator-level isErrored handling and the callback from the generic decorated JSON loader.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in `c89c61c`. The old result, join, scan, virtual-table, and temp tags now resolve through legacyNodeTags to current rendering entries. Only early-probe retains an override because its builder crosslink has no current equivalent.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in `c89c61c`. Kept the join, scan, and expression grouping comments in the legacy alias list.

query-graphs/test/hyper.test.ts
vogelsgesang: 🤖 Addressed in `c89c61c`. Removed the legacy-plan test case.

query-graphs/test/hyper.test.ts
vogelsgesang: remove the `error` and `running`, as they are no longer present in new plans

query-graphs/src/loaders/hyper.ts
```diff
@@ -77,36 +57,68 @@ const nodeRenderingConfig: Record<string, NodeRenderingConfig> = {
     "exp:reference": {displayNameKey: "id"},
 };
 
-// Legacy tags from before the kebab-case transition.
+// TODO(2026-12-18): Remove the early-probe override together with the legacy tag aliases below.
+const legacyNodeRenderingConfig: Record<string, NodeRenderingConfig> = {
+    "op:early-probe": {icon: "filter-symbol", crosslinkSourceKey: "builder"},
+};
```
vogelsgesang: remove this already today; early-probe is uncommon enough; also this allows us to avoid adding handling for `legacyNodeRenderingConfig`

query-graphs/src/loaders/hyper.ts
vogelsgesang: afaik, there are still "direct tree" plans, e.g. the non-analyzed queries?

query-graphs/test/hyper.test.ts
vogelsgesang: 🤖 Addressed in `dfea5a8`. Removed the obsolete operator-level `error` and `running` fields and the corresponding metadata assertion.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 The refreshed non-analyzed fixtures are also `{tree, pipelines}` envelopes (including insert, metadata, table construction/function, and TPC-H Q2). Optimizer-step output remains a separate supported wrapper. The dated TODO only targets bare `isHyperPlanRoot` plans emitted before 2026-09-18; it does not target optimizer steps.

query-graphs/src/loaders/hyper.ts
vogelsgesang: 🤖 Addressed in `452ecec`. Removed `early-probe` support and the now-unnecessary `legacyNodeRenderingConfig` lookup path.

query-graphs/src/loaders/hyper.ts
```diff
@@ -144,10 +150,8 @@ const hyperConfig: DecoratedJsonTreeConfig = {
         // Expand expression details by default.
         return nodeTypeKey !== "operator";
     },
-    isErrored(rawNode, metadata) {
-        return metadata.has("Error") && tryGetPropertyPath(rawNode, ["statistics", "running"]) === true;
-    },
     getNodeColorValue(rawNode) {
+        // TODO(2026-12-18): Remove operator-level CPU cycles after the pipeline-statistics compatibility window.
         const executionTime = tryGetPropertyPath(rawNode, ["statistics", "cpu-cycles"]);
```
vogelsgesang: for new Hyper planes, I still want to highlight the most expensive pipeline driver in terms of cpu-cycles. Do it via `applyPipelineStatistics`. In general, I think we should remove the `getNodeColorValue` and instead use post-processing passes to add color highlighting.

vogelsgesang: 🤖 Addressed in c63f8da. `applyPipelineStatistics` now colors pipeline drivers by their relative `cpu-cycles`. I also removed the generic `getNodeColorValue` conversion hook and moved DuckDB and legacy Hyper coloring into loader-specific post-processing passes.

PR #186, opened 2026-09-19
Show plan text and SQL text

vogelsgesang:
With this comment, we now also render the JSON / XML plan in its textual representation inside a collapsible panel.

The `TreeDescription` can now include `TextDocument`s. The general loader infrastructure always makes the JSON / XML plan available in that way.

Furthermore, plan readers can extract, e.g., SQL text from the query plan. The DuckDB plans indeed embed the SQL query, and the DuckDB loader makes the query available this way.

For now, we use a simple unstyled read-only text area. In the future, we could also integrate, e.g., syntax highlighting.

We also offer a "Copy" button to copy the text documents to the clipboard. For that purpose, a reusable CopyButton provides clipboard progress, success, and failure feedback with accessible announcements.


query-graphs/src/tree-description.ts
```diff
@@ -61,13 +61,22 @@ export interface Crosslink {
...
     /// The tree root
     root: TreeNode;
     /// Metadata about the graph; displayed in the top-level tree label
     metadata?: Map<string, string>;
     /// Whether the metadata panel should be visually highlighted
     metadataHighlighted?: boolean;
+    /// Text documents associated with the graph
+    textDocuments?: TextDocument[];
     /// Additional links between indirectly related nodes
     crosslinks?: Crosslink[];
```
vogelsgesang: cross links first. Here and all users

```suggestion
    /// Additional links between indirectly related nodes
    crosslinks?: Crosslink[];
    /// Text documents associated with the graph
    textDocuments?: TextDocument[];
```

query-graphs/src/tree-description.ts
```diff
@@ -61,13 +61,22 @@ export interface Crosslink {
     target: TreeNode;
 }
 
+export interface TextDocument {
+    id: string;
+    title: string;
+    text: string;
+    language?: string;
```
vogelsgesang, line 68: should we use mime-type instead? or file extension? Would either help somehow for the copy gesture to provide more metadata?

query-graphs/src/loaders/index.ts
```diff
@@ -54,6 +54,17 @@ function stripSurroundingText(text: string): string {
...
+function addPlanDocument(plan: LoadedPlan, text: string, language: string): LoadedPlan {
+    const planDocument: TextDocument = {id: "plan", title: "Query Plan", text, language};
+    return {
+        ...plan,
+        tree: {
+            ...plan.tree,
+            textDocuments: [...(plan.tree.textDocuments ?? []), planDocument],
+        },
+    };
+}
```
vogelsgesang, line 62: can't we just modify `plan` in-place instead of recreating it?

query-graphs/src/loaders/index.ts
```diff
@@ -69,7 +80,7 @@ export function loadPlanFromText(text: string, options: LoadPlanOptions = {}): L
         try {
             const json = JSON.parse(planText) as Json;
             const plan = loadMatchingPlan(json, jsonPlanLoaders, errors, format);
-            if (plan !== undefined) return plan;
+            if (plan !== undefined) return addPlanDocument(plan, text, "json");
```
vogelsgesang: `text` itself is not necessarily JSON / XML in case we stripped something.
We should probably use the `planText` instead.

query-graphs/test/duckdb.test.ts
```diff
@@ -163,6 +165,7 @@ test("DuckDB resolves analyzed delimiter targets from operator types", () => {
     const source = tree.root.children?.[0];
 
     assert.deepEqual(tree.crosslinks, [{source, target: tree.root}]);
+    assert.deepEqual(tree.textDocuments, [{id: "query", title: "Original SQL Query", text: query, language: "sql"}]);
```
vogelsgesang: remove `query_name` and this assertion from this test. It's unrelated to `query_name`

query-graphs/src/tree-description.ts
vogelsgesang: 🤖 Addressed

query-graphs/src/tree-description.ts
vogelsgesang: 🤖 Kept `language`: `navigator.clipboard.writeText()` publishes plain text and cannot attach filename or extension metadata. A MIME type would only affect copying if we later switched to `ClipboardItem`; the language hint remains useful for future rendering such as syntax highlighting.

query-graphs/src/loaders/index.ts
vogelsgesang: 🤖 Addressed

query-graphs/src/loaders/index.ts
vogelsgesang: 🤖 Addressed

query-graphs/test/duckdb.test.ts
vogelsgesang: 🤖 Addressed

PR #187, opened 2026-09-19
Keep plan-dumper config out of explained SQL

vogelsgesang:
The plan dumper currently sends its configuration comments to databases as part of the explained SQL. DuckDB consequently embeds those comments in its query text, while source positions in other plan formats refer to the configuration prefix rather than the SQL users wrote.

Reserve triple-dash comments for plan-dumper configuration and strip them before sending queries to a database. Ordinary SQL comments remain intact, while a shared formatter inserts EXPLAIN after the leading comment block and puts the query on the following line for every supported database. Comments explaining UNSUPPORTED directives are configuration comments too, keeping database-specific limitations out of rendered SQL.

Regenerated fixtures now reflect the new input across all engines. DuckDB exposes the cleaned embedded SQL, Hyper sqlpos offsets and Umbra sourceLocation values track the reformatted statement, and loader golden hashes are refreshed. PostgreSQL, MariaDB, and current CedarDB v2026-09-16 produced no substantive plan changes after volatile runtime data was filtered; the recorded PostgreSQL version advances to 17.11.


plan-dumper/queries/tpch/tpch-q15.sql
```diff
@@ -1,4 +1,4 @@
--- UNSUPPORTED: cedardb
+--- UNSUPPORTED: cedardb
 -- CedarDB v2026-08-13 hits an internal Pipeline::getRawLocalStatePtr assertion
 -- during EXPLAIN ANALYZE; plain EXPLAIN succeeds.
```
vogelsgesang: should be a configuration comment, given it explains the UNSUPPORTED

Here and other similar files

query-graphs/test/loader-results.json
vogelsgesang, line 1: I am surprised none of the source locations in the other databases changed with this change?

plan-dumper/queries/tpch/tpch-q15.sql
vogelsgesang: 🤖 Addressed

query-graphs/test/loader-results.json
vogelsgesang: 🤖 Only the DuckDB fixtures were regenerated in this PR because the reported issue is their embedded query text. The other engines' stored plans were left untouched, so their source locations and golden hashes remain unchanged. A future regeneration with the updated dumper may shift those locations due to the new EXPLAIN placement.

vogelsgesang: 🤖 Follow-up: all engines are now regenerated. Umbra `sourceLocation` values and Hyper `sqlpos` offsets reflect the reformatted statements; PostgreSQL and MariaDB have no corresponding source-position changes, and the verified latest CedarDB v2026-09-16 produced no substantive net plan diff. Volatile runtime data was filtered and the loader golden hashes were refreshed.

PR #188, opened 2026-09-19
Add an interactive JSON plan viewer

vogelsgesang:
## Summary

- replace the JSON plan textarea with a read-only CodeMirror 6 viewer
- add syntax highlighting, line numbers, a folding gutter, and keyboard folding
- load CodeMirror only when the panel is first opened
- content-hash and precache async chunks so the viewer remains available offline
- retain the lightweight textarea for non-JSON text documents

The source text remains unchanged. As a result, CodeMirror offers gutter folding for multiline JSON structures; compact single-line JSON has no line-based fold points.

## Verification

- `pnpm exec tsc --noEmit --incremental false -p standalone-app/tsconfig.json`
- `pnpm --filter @tableau/query-graphs build`
- `pnpm --filter query-graphs-app build`
- `pnpm test` (46 tests)
- `pnpm lint-test`
- verified generated service worker precaches both content-hashed CodeMirror chunks


standalone-app/src/TreeLabel.tsx
```diff
@@ -38,22 +67,24 @@ export function TreeLabel({title, setTitle, metadata, metadataHighlighted, textD
...
             />
-            {metadataChildren.length > 0 ? (
-                <CollapsiblePanel title="Plan Metadata" highlighted={metadataHighlighted}>
-                    <div className="graph-metadata">{metadataChildren}</div>
-                </CollapsiblePanel>
-            ) : null}
-            {textDocuments?.map((document) => (
-                <TextDocumentPanel key={document.id} document={document} />
-            ))}
+            <div className="graph-sidebar-panels">
```
vogelsgesang: generalize to `graph-sidebar-scrollable`?

standalone-app/src/TreeLabel.tsx
```diff
@@ -1,9 +1,13 @@
-import type {ReactElement} from "react";
+import {lazy, Suspense, useState, type ReactElement} from "react";
 import {CollapsiblePanel} from "@tableau/query-graphs/lib/ui/CollapsiblePanel";
 import {CopyButton} from "@tableau/query-graphs/lib/ui/CopyButton";
 import type {TextDocument} from "@tableau/query-graphs/lib/tree-description";
 import "./TreeLabel.css";
 
+const JsonDocument = lazy(() => import("./JsonDocument").then((module) => ({default: module.JsonDocument})));
+const SqlDocument = lazy(() => import("./SqlDocument").then((module) => ({default: module.SqlDocument})));
+const CodeDocument = lazy(() => import("./CodeDocument").then((module) => ({default: module.CodeDocument})));
```
vogelsgesang:
```suggestion
// Lazy-load the editors, so we don't need to load CodeMirror and it's plugins as part of the
// initial page load. Those bundles are XXX bytes vs. XXX bytes of the remaining main
// application, hence it's worth to lazy-load them.
const JsonDocument = lazy(() => import("./JsonDocument").then((module) => ({default: module.JsonDocument})));
const SqlDocument = lazy(() => import("./SqlDocument").then((module) => ({default: module.SqlDocument})));
const CodeDocument = lazy(() => import("./CodeDocument").then((module) => ({default: module.CodeDocument})));
```

standalone-app/src/TreeLabel.tsx
```diff
@@ -12,17 +16,42 @@ export interface TreeLabelProps {
     textDocuments?: TextDocument[];
 }
 
+function LoadingDocument({title}: {title: string}) {
```
vogelsgesang: move to separate `LoadingDocument` file; also move relevant css into separate file

standalone-app/src/TreeLabel.css
```diff
@@ -31,4 +45,84 @@
...
+    & span:nth-child(4) {
+        width: 76%;
+    }
+
+    & span:nth-child(5) {
+        width: 84%;
+    }
+}
+
+.graph-sidebar .qg-collapsible-panel-content:has(> .graph-text-document) {
```
vogelsgesang: why is this needed?

vogelsgesang: 🤖 Removed this override. The CodeMirror editor already owns its internal scrolling, while the collapsible panel's existing overflow behavior safely handles unusually short viewports; forcing `overflow: hidden` here was redundant and could clip the editor.

query-graphs/src/ui/CollapsiblePanel.tsx
```diff
@@ -1,21 +1,59 @@
-import type {ReactNode} from "react";
+import {useRef, useState, type ReactNode} from "react";
 import cc from "classcat";
 import "./CollapsiblePanel.css";
 
 export interface CollapsiblePanelProps {
+    /** Content displayed in the panel summary. */
     title: ReactNode;
+    /** Controls displayed at the end of the summary without toggling the panel. */
```
vogelsgesang:
```suggestion
    /** Controls displayed at the end of the header line; available also without toggling the panel. */
```

query-graphs/src/ui/CollapsiblePanel.tsx
```diff
@@ -1,21 +1,59 @@
-import type {ReactNode} from "react";
+import {useRef, useState, type ReactNode} from "react";
 import cc from "classcat";
 import "./CollapsiblePanel.css";
 
 export interface CollapsiblePanelProps {
+    /** Content displayed in the panel summary. */
```
vogelsgesang:
```suggestion
    /** Content displayed in the header. */
```

query-graphs/src/ui/CollapsiblePanel.tsx
```diff
@@ -31,7 +59,7 @@ export function CollapsiblePanel({title, headerActions, children, highlighted, c
                     </span>
                 ) : null}
             </summary>
-            <div className="qg-collapsible-panel-content">{children}</div>
+            {!mountContentOnFirstIntent || contentMounted ? <div className="qg-collapsible-panel-content">{children}</div> : null}
```
vogelsgesang, line 62: should we instead pass `mountContentOnFirstIntent` as parameter to `useState` above?

standalone-app/src/CodeMirrorDocument.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const path = document.createElementNS("http://www.w3.org/2000/svg", "path");
+    path.setAttribute("d", "M5 3.5 10 8l-5 4.5");
+    svg.append(path);
+    marker.append(svg);
+    return marker;
+}
+
+const folding = foldGutter({markerDOM: createFoldMarker});
+const noLanguageExtension: Extension = [];
+const foldMarkerTheme = EditorView.baseTheme({
```
vogelsgesang, line 25: keep the folding code close together and don't intermix the unrelated `noLanguageExtension` in betwee

standalone-app/src/CodeMirrorDocument.tsx
```diff
@@ -0,0 +1,98 @@
...
+        stroke: "currentColor",
+        strokeWidth: "1.75",
+        strokeLinecap: "round",
+        strokeLinejoin: "round",
+    },
+    ".graph-fold-marker-open svg": {
+        transform: "rotate(90deg)",
+    },
+});
+const documentTheme = EditorView.theme({
```
vogelsgesang, line 49: why do we have this here instead of in the CSS file?

standalone-app/src/CodeMirrorDocument.tsx
```diff
@@ -0,0 +1,98 @@
...
+                    keymap.of([...defaultKeymap, ...foldKeymap]),
+                    EditorState.readOnly.of(true),
+                    EditorView.contentAttributes.of({"aria-label": document.title}),
+                ],
+            }),
+        });
+        return () => view.destroy();
+    }, [document, languageExtension]);
+
+    return <div ref={parent} className="graph-text-document nowheel nodrag nopan" />;
```
vogelsgesang: `nowheel nodrag nopan` - elaborate on each one of them why we need it

standalone-app/src/TreeLabel.css
```diff
@@ -1,34 +1,139 @@
 .graph-sidebar {
+    --graph-sidebar-paint-gutter: 10px;
+    --graph-sidebar-outer-margin: 5px;
```
vogelsgesang: is it really necessary to split this? Should we just use only `--graph-sidebar-paint-gutter` of 15px and no outer margin?

query-graphs/src/ui/CollapsiblePanel.tsx
vogelsgesang: 🤖 Addressed.

query-graphs/src/ui/CollapsiblePanel.tsx
vogelsgesang: 🤖 Addressed.

query-graphs/src/ui/CollapsiblePanel.tsx
vogelsgesang: 🤖 Kept `contentMounted` initialized to `false`: it records whether intent has occurred, while `mountContentOnFirstIntent` independently controls whether deferral is enabled. Passing the prop to `useState` would derive state only on the first render and conflate those two concerns.

standalone-app/src/CodeMirrorDocument.tsx
vogelsgesang: 🤖 Addressed.

standalone-app/src/CodeMirrorDocument.tsx
vogelsgesang: 🤖 Kept this as an `EditorView.theme` and documented why: CodeMirror injects scoped base styles at runtime, including its default focus outline. The theme extension gives our root/scroller overrides predictable precedence without specificity hacks or `!important`; ordinary wrapper sizing remains in `CodeMirrorDocument.css`.

standalone-app/src/CodeMirrorDocument.tsx
vogelsgesang: 🤖 Moved the React Flow interaction guards to the surrounding sidebar, where the integration belongs. `nowheel` lets documents scroll without zooming the canvas; `nopan` lets users select text and use controls without dragging the canvas. Removed `nodrag` because the sidebar is not a draggable node.

standalone-app/src/TreeLabel.css
vogelsgesang: 🤖 Addressed. The sidebar now uses one 15px paint gutter with zero outer margin; viewport width and height constraints are correspondingly simpler while visible content remains at the same 15px inset.

standalone-app/src/TreeLabel.tsx
```diff
@@ -1,9 +1,13 @@
-import type {ReactElement} from "react";
+import {lazy, Suspense, type ReactElement} from "react";
 import {CollapsiblePanel} from "@tableau/query-graphs/lib/ui/CollapsiblePanel";
 import {CopyButton} from "@tableau/query-graphs/lib/ui/CopyButton";
 import type {TextDocument} from "@tableau/query-graphs/lib/tree-description";
 import "./TreeLabel.css";
 
+const loadDocumentPane = () =>
+    import(/* webpackChunkName: "editor" */ "./DocumentPane").then((module) => ({default: module.DocumentPane}));
+const DocumentPane = lazy(loadDocumentPane);
```
vogelsgesang: we no longer need the `loadDocumentPane` indirection

